PR #25, opened 2023-08-28
feat: using Node.js v18 for development

chenjiahan:
## Summary

- Using Node.js v18 for development
- Run `test-macOS.yml` in both Node 16 / 18 to keep compatible with Node 16.
- Remove `monorepo.code-workspace` file because it is useless.
- Add corepack config to package.json.

## Checklist


sanyuan0704:
I will check the e2e problem.

PR #34, opened 2023-08-30
feat: beginner for text annotation

bvanjoi:
https://github.com/web-infra-dev/rspress/assets/30187863/088b6968-0b4b-4c49-b12c-3a390bb0824e


packages/plugin-annotation/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@rspress/plugin-annotation",
+  "version": "0.0.6",
```
sanyuan0704, line 3: The version should be updated to latest v0.0.10

packages/plugin-annotation/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import path from 'path';
+import type { RspressPlugin } from '@rspress/shared';
+
+const annotationComponent = path.resolve(__dirname, '../static/index.tsx');
+
+export function pluginAnnotation(): RspressPlugin {
+  return {
+    name: '@modern-js/doc-plugin-annotation',
```
sanyuan0704, line 8: @rspress/plugin-annotation

packages/plugin-annotation/src/index.ts
```diff
@@ -0,0 +1,11 @@
+import path from 'path';
+import type { RspressPlugin } from '@rspress/shared';
+
+const annotationComponent = path.resolve(__dirname, '../static/index.tsx');
```
sanyuan0704, line 4: Should think of windows path, refer https://github.com/web-infra-dev/rspress/blob/main/packages/plugin-medium-zoom/src/index.ts

packages/core/src/theme-default/layout/DocLayout/index.tsx
```diff
@@ -83,12 +83,14 @@ export function DocLayout(props: DocLayoutProps) {
           {isOverviewPage ? (
             <Overview />
           ) : (
-            <div className="rspress-doc">
-              <TabDataContext.Provider value={{ tabData, setTabData }}>
-                <MDXProvider components={getCustomMDXComponent()}>
-                  <Content />
-                </MDXProvider>
-              </TabDataContext.Provider>
+            <div>
```
sanyuan0704, line 86: Why is there an extra container div?

sanyuan0704:
The e2e test case seems failed.

packages/core/src/theme-default/layout/DocLayout/index.tsx
bvanjoi: We've removed all child elements of the `﻿.rspress-doc`, and I think the ﻿footer should not be considered part of the article. Thus, I add a `div` to keep `<DocFooter/>`

see: https://github.com/web-infra-dev/rspress/pull/34/files#diff-87b592124a448b2d70388b0cfb70aa567357fbb6140d9de24df2ed6ba5440778R19-R21

chenjiahan:
This PR is outdated so I'm closing it

PR #59, opened 2023-09-06
docs: change doc-plugin-typedoc to plugin-typedoc

evanryuu:
## Summary

1. [Official Document] Change `doc-plugin-typedoc` to `plugin-typedoc`
2. [plugin-typedoc] Change the name of the plugin

## Details

I notice that the guide for `@rspress/doc-plugin-typedoc` seems not correct since there's only `@rspress/plugin-typedoc` package on npmjs.org, so I update the docs a little.

Besides the change of official document, I also change the index file of `@rspress/plugin-typedoc` to make it consistent with other plugins.
image

image

But I did not run `pnpm run change` since I believe this should be decided by the maintainers 😂

## Checklist

### DIFF
--- a/packages/document/docs/en/plugin/official-plugins/overview.mdx
+++ b/packages/document/docs/en/plugin/official-plugins/overview.mdx
@@ -2,4 +2,4 @@
 
 Official plugins include the following:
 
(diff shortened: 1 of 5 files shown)


sanyuan0704:
Yeah, you are right.Thank you for correcting it.

sanyuan0704 (approved):

PR #108, opened 2023-09-22
feat: add min-height in .rspress-doc

Timeless0911:
## Summary

copilot:summary

## Details

copilot:walkthrough

## Checklist


packages/core/src/theme-default/layout/DocLayout/index.module.scss
```diff
@@ -14,6 +14,10 @@
   display: none;
 }
 
+:global(.rspress-doc) {
+  min-height: 600px;
```
chenjiahan, line 18: How about `xx vh`, which is more friendly on small screens.

Timeless0911: use 60 vh instead

chenjiahan (dismissed):

chenjiahan (approved):

Issue #93, opened 2023-09-19
[Feature]: Remove '.html' suffix of url links


xalanq:
### What problem does this feature solve?

![image](https://github.com/web-infra-dev/rspress/assets/30496124/7c96a1f8-3b7b-4bed-8c2f-9f0091b28f2e)

### What does the proposed API look like?

nothing

sanyuan0704:
Why do you want to remove the suffix?

xalanq:
> Why do you want to remove the suffix?

Make urls shorter and look better

sanyuan0704:
SSG output are html files, so bring .html to make the path more accurate and intuitive

xalanq:
> SSG output are html files, so bring .html to make the path more accurate and intuitive

That's no doubt that the output files are html files. And I could also write some nginx rules to remove the '.html' suffix from the url.

BUT what I want is that the urls generated by rspress on `sidebar` and `navbar` should clean the '.html' suffix. Becoz It's so weird that there are still the '.html' suffix with the urls when user hover the `<a>` tag links in 2023, 21st century.

sanyuan0704:
I think it's worth sacrificing URL length for compatibility. After removing the .html suffix, if you jump to a link and want to share the link with others, they may not be able to access the HTML file because there is no .html suffix.

chenjiahan:
Vuejs: https://vuejs.org/api/

![image](https://github.com/web-infra-dev/rspress/assets/7237365/3ca92127-ece6-45c1-9bca-7c0588c6bdc2)

![image](https://github.com/web-infra-dev/rspress/assets/7237365/2e00895c-6b00-426c-88e2-698eb9891099)

xalanq:
> Vuejs: https://vuejs.org/api/
> 
> ![image](https://user-images.githubusercontent.com/7237365/269184434-3ca92127-ece6-45c1-9bca-7c0588c6bdc2.png)
> 
> ![image](https://user-images.githubusercontent.com/7237365/269184206-2e00895c-6b00-426c-88e2-698eb9891099.png)

- React：https://react.dev/learn/installation
![image](https://github.com/web-infra-dev/rspress/assets/30496124/d0529351-016a-4cbf-9de1-332656aba1f3)

- GitBook: e.g. https://react.i18next.com/
![image](https://github.com/web-infra-dev/rspress/assets/30496124/de6f4328-ef82-4521-aa97-3206b4b16699)

[...]

PR #121, opened 2023-09-26
feat: support cleanUrls to generate urls without .html

zoolsher:
## Summary

Support a new config as `route.cleanUrls` to remove the generated links suffix like `.html`

## Related Issue

[#93](https://github.com/web-infra-dev/rspress/issues/93)

## Checklist


zoolsher:
@sanyuan0704 Do you have any suggestion on this feature?

packages/core/src/theme-default/components/Overview/index.tsx
```diff
@@ -26,6 +26,7 @@ export function Overview() {
     siteData,
     page: { routePath },
   } = usePageData();
+  const cleanUrls = !!usePageData()?.siteData?.route?.cleanUrls;
```
sanyuan0704, line 29: `siteData` has beed destructed before so there is no need to call `usePageData`. 
```ts
const cleanUrls = siteData.route?.cleanUrls
```

packages/core/src/runtime/utils.ts
```diff
@@ -23,14 +24,18 @@ export function removeBase(url: string): string {
...
 }
 
-export function isEqualPath(a: string, b: string) {
-  return withBase(normalizeHref(a)) === withBase(normalizeHref(b));
+export function isEqualPath(a: string, b: string, cleanUrls: boolean) {
+  return withBase(normalizeHref(a, cleanUrls)) === withBase(normalizeHref(b, cleanUrls));
+}
+
+export function useNormalizeHrefInRuntime(a: string){
+  const cleanUrls = !!usePageData()?.siteData?.route?.cleanUrls;
```
sanyuan0704, line 32: It's better to use `Boolean()` instead of `!!` to avoid ambiguity and some of optional chain is not neccessary, such as:

```ts
const cleanUrls = Boolean(usePageData().siteData.route?.cleanUrls);
```

sanyuan0704:
> @sanyuan0704 Do you have any suggestion on this feature?

Yeah, i'm happy to review this pr.

packages/core/src/runtime/utils.ts
```diff
@@ -23,14 +24,18 @@ export function removeBase(url: string): string {
   return rawRemoveBase(url, siteData.base);
 }
 
-export function isEqualPath(a: string, b: string) {
-  return withBase(normalizeHref(a)) === withBase(normalizeHref(b));
+export function isEqualPath(a: string, b: string, cleanUrls: boolean) {
+  return withBase(normalizeHref(a, cleanUrls)) === withBase(normalizeHref(b, cleanUrls));
+}
+
+export function useNormalizeHrefInRuntime(a: string){
```
sanyuan0704, line 31: I think it's better to implement the normalize ability by a normal util function instead of a react hook.The reasons are as follows:

1. We don't need to call the `usePageData` hook because we can get siteData from `virtual-site-data` directly.

2. The normalize process is a high frequency operation in all kinds of components.The react hook has a limitation, that is, the hook cannot be placed in a conditional statement to ensure that the hook execution order is always consistent, otherwise there will be some rendering bugs.We can avoid the problem as much as possible.

3. The name of utils can still be `normalizeHref` to make it clear and easier to understand. We can just import it from `@/runtime` instead of `@rspress/shared`.

zoolsher: I understand the first two points and have already made the necessary adjustments.

Regarding the third point, I comprehend your perspective, but I have a different opinion on it. My concern is that having two identical symbols in the codebase could lead to confusion. For instance, when someone types `normalizeHref` in a runtime component code, there is a possibility that VSCode automatically imports the symbol from either `@/runtime` or `@rspress/shared`. This could result in unexpected behavior and make it difficult to locate any bugs.

Therefore, I suggest creating two separate symbols for each function to avoid these potential issues.

packages/core/src/runtime/utils.ts
zoolsher: Fixed it.

packages/core/src/global.d.ts
```diff
@@ -11,10 +11,10 @@ declare module 'virtual-routes-ssr' {
 }
 
 declare module 'virtual-site-data' {
-  import { SiteData } from 'shared/types';
```
zoolsher, line 14: @sanyuan0704 can you help review this change? I'm not sure whether this change matches your expectation

sanyuan0704: The change is useful and solve the type hint problem for siteData.

packages/core/src/runtime/utils.ts
sanyuan0704: I agree with you.And the `useNormalizeHrefInRuntime` could be renamed to `normalizeHrefInRuntime` because it's no longer a react hook.

zoolsher:
@chenjiahan @sanyuan0704 Please help review the preview deploy docsite in case I miss something.

chenjiahan:
LGTM 👍

packages/document/docs/zh/api/config/config-basic.mdx
```diff
@@ -310,3 +310,20 @@ export default defineConfig({
   },
 });
 ```
+
+### route.cleanUrls
+
+- Type: `Boolean`
+- Default: `false`
+
+开启后可以生成无 `.html` 后缀的链接，URL 可以更加简洁
```
sanyuan0704, line 319: There missed a period.

.changeset/nice-zebras-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+'@rspress/shared': minor
+'@rspress/core': minor
+---
+
+support cleanUrls to generate url without suffix
```
sanyuan0704, line 6: The changeset info should be like this format:
```
feat: support cleanUrls to generate url without suffix

feat: 支持 `cleanUrls` 参数以去掉 url 的后缀名
```

chenjiahan: The Chinese message should be optional.

Maybe Rspress do not need a Chinese changelog (or generate by GPT)

sanyuan0704 (approved):

Issue #123, opened 2023-09-27
[Feature]: Improve the transition when navigating to an URL with anchor


chenjiahan:
### What problem does this feature solve?

When navigating to an URL with anchor, the page will move up and down once, which looks a bit strange:

https://github.com/web-infra-dev/rspress/assets/7237365/3be5e9a3-cb62-4ba6-9d41-cb41f7af8f46

Reproduce steps:

- Open https://deploy-preview-121--aquamarine-blini-95325f.netlify.app/api/config/config-build#builderconfig
- Click "builderConfig" link.

### What does the proposed API look like?

Provide a more natural transition effects.

zoolsher:
I'll try to fix this

PR #125, opened 2023-09-27
fix: not play animation when first entry

zoolsher:
## Summary

Don't call the startViewTransition when transforming between the same display name (from route path) component.

In the original issue(#123), @chenjiahan reported a behavior where there are two animations when jumping to a page with a hash. The reasons for this happening are:

1. The page jumps within the markdown do not use <Link/>, but <a/>, which means the browser will reload when someone clicks on the link.
2. The view transition animation is played even when jumping within the same page (same route path).

Fixing the second issue is quite easy. However, it becomes challenging when trying to fix the first one.

I was working on the navigation between pages with a hash, for example, from `guide/start` to `config/repress#nav`. I needed some time to dispatch an event so that the `sideEffect.ts` would know that the DOM is attached to the browser. Otherwise, the scroll position and animation would be incorrect. To find the logic that handles this event, you can search for `RspressReloadContent` in this codebase. I decided not to include the logic in the `useViewTransition` hooks because I wanted the handling of scroll and position to be in one place to avoid confusion.

## Details

- [x] Not paly animation when first entry.
- [x] Link component within the same site just should use push state rather than href.

## Related Issue

#123 

## Checklist

- [x] I have updated the documentation. // There is no need to update the documentation.

### DIFF
--- a/.changeset/curvy-buses-joke.md
+++ b/.changeset/curvy-buses-joke.md
@@ -1,0 +1,6 @@
+---
+'@rspress/docs': minor
+'@rspress/core': minor
+---
+
+fix: tuning the animation for `first entry` and `navigate with hash`
--- a/e2e/fixtures/view-transition/doc/_meta.json
+++ b/e2e/fixtures/view-transition/doc/_meta.json
@@ -1,0 +1,10 @@
+[
+  {
+    "text": "guide",
+    "link": "/guide"
+  },
+  {
+    "text": "start",
+    "link": "/start"
+  }
+]
--- a/e2e/tests/view-transition.test.ts
+++ b/e2e/tests/view-transition.test.ts
@@ -35,10 +35,13 @@ test.describe('basic test', async () => {
     await session.send('Animation.enable');
     const [end, resolve] = getPendingPromise();
     session.on('Animation.animationStarted', payload => {
-      expect(payload.animation.type).toBe('CSSAnimation');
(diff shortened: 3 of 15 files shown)


zoolsher:
@sanyuan0704 I need your help on this.

sanyuan0704:
> @sanyuan0704 I need your help on this. Would you mind reviewing the change in `packages/core/src/theme-default/logic/sideEffects.ts`

ok

zoolsher:
@chenjiahan Would you mind checking this? Would love to know your opinion!

zoolsher:
> > @sanyuan0704 I need your help on this. Would you mind reviewing the change in `packages/core/src/theme-default/logic/sideEffects.ts`
> 
> ok

I added more context on the top

chenjiahan:
The transition effects between pages are now very smooth. 😄

However, when I click on an anchor within a page, the anchor does not seem to work. Here is a case:

1. open https://deploy-preview-125--aquamarine-blini-95325f.netlify.app/guide/advanced/custom-theme.html
2. click "Redevelop a custom theme"



https://github.com/web-infra-dev/rspress/assets/7237365/eaaf0266-4e4d-49ae-9f23-09b2c935cc72

sanyuan0704:
@chenjiahan Now visit https://deploy-preview-125--aquamarine-blini-95325f.netlify.app/guide/advanced/custom-theme and the hash link effect is normal

sanyuan0704 (approved):

chenjiahan:
LGTM👍🏻

PR #128, opened 2023-09-28
fix: dead link invalid when having a trailing slash

### LINKED ISSUE #126: [Bug]: path without /index will be recognized as dead links
### Version

```sh
System:
    OS: macOS 12.5.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 70.20 MB / 32.00 GB
    Shell: 5.8.1 - /bin/zsh
  Browsers:
    Chrome: 117.0.5938.92
    Safari: 15.6.1
  npmPackages:
    @rspress/shared: 0.2.0 => 0.0.0-next-20230927072732
```

### Details

use 0.2.0 next version

It used to be written as `/en/api/#config`, but now the dead link check will report an error, and it needs to be written as `/en/api/index#config`

image

image

### Reproduce link

none

### Reproduce Steps

none

sanyuan0704:
## Summary

Fix #126 

## Details

The `normalizeRotuePath` needs to handle the path which has a trailing slash.

## Checklist

### DIFF
--- a/packages/core/src/node/route/RouteService.ts
+++ b/packages/core/src/node/route/RouteService.ts
@@ -31,6 +31,7 @@ export const normalizeRoutePath = (
   base: string,
   version: string,
 ): { routePath: string; lang: string; version: string } => {
+  const hasTrailSlash = routePath.endsWith('/');
   let versionPart = '';
   let langPart = '';
   let purePathPart = '';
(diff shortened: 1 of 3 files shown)


packages/core/src/node/route/RouteService.ts
```diff
@@ -51,13 +52,18 @@ export const normalizeRoutePath = (
...
-  const normalizedRoutePath = addLeadingSlash(
+  let normalizedRoutePath = addLeadingSlash(
     [versionPart, langPart, purePathPart].filter(Boolean).join('/'),
   )
     // remove the extension
     .replace(/\.[^.]+$/, '')
     .replace(/\/index$/, '/');
 
+  // restore the trail slash
+  if (hasTrailSlash && !normalizedRoutePath.endsWith('/')) {
```
chenjiahan, line 63: can we add some test cases for this

sanyuan0704: done

chenjiahan (approved):

PR #144, opened 2023-10-08
feat: use jiti to load config

chenjiahan:
## Summary

- Less dependencies (jiti is used by tailwindcss v3.3)
- Faster startup time
- Better support for monorepo

Screenshot 2023-10-08 at 10 38 58

Screenshot 2023-10-08 at 10 39 44

## Related Issue

https://github.com/web-infra-dev/modern.js/pull/4754

## Checklist

### DIFF
--- a/.changeset/gorgeous-numbers-nail.md
+++ b/.changeset/gorgeous-numbers-nail.md
@@ -1,0 +1,7 @@
+---
+'rspress': patch
+---
+
+feat: use jiti to load config
+
+feat: 使用 jiti 来加载配置文件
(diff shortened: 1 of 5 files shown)


sanyuan0704:
The unit test in cli package seems failed after switching to jiti.

chenjiahan:
Yes, there is a ESM/CJS interop issue and I will try to figure it out

chenjiahan:
Jiti uses CommonJS transforms and this will break some dynamic imports from Rspress.

It looks like Jiti v2 may fix these issues,  ref: https://github.com/unjs/jiti/issues/32

This PR will be held until Jiti fixes it or we find some alternatives.

PR #182, opened 2023-10-13
style: Logo icon adaptation.

youbudong:
Change the size of the icon to adapt based on the height instead of the width, as the length of the logo text varies slightly.


sanyuan0704:
Thank you!

sanyuan0704 (approved):

PR #191, opened 2023-10-16
fix: modernjs single language error

sanyuan0704:
## Summary

copilot:summary

## Details

copilot:walkthrough

## Checklist


packages/modern-plugin-rspress/src/features/lanchDoc.ts
```diff
@@ -129,6 +129,7 @@ export async function launchDoc({
         locales,
       },
       markdown: {
+        mdxRs: false,
```
10Derozan, line 132: can remove this line ? we can unable it  in plugin which unsupport mdxRs

sanyuan0704: Currently we disabled it in all the places. Once the mdx-rs support relavant plugins, we will store the config.

PR #194, opened 2023-10-16
fix(style): limit the max width of hero tagline

chenjiahan:
## Summary

Limit the max width of hero tagline.

Before:

download

After:

FireShot Capture 056 - Rsbuild - localhost

## Checklist

### DIFF
--- a/packages/core/src/theme-default/components/HomeHero/index.tsx
+++ b/packages/core/src/theme-default/components/HomeHero/index.tsx
@@ -17,6 +17,9 @@ const DEFAULT_HERO = {
 export function HomeHero({ frontmatter }: { frontmatter: FrontMatterMeta }) {
   const hero = frontmatter?.hero || DEFAULT_HERO;
   const hasImage = hero.image !== undefined;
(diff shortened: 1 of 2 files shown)


packages/core/src/theme-default/components/HomeHero/index.tsx
```diff
@@ -34,16 +37,16 @@ export function HomeHero({ frontmatter }: { frontmatter: FrontMatterMeta }) {
           </h1>
           {hero.text?.length && (
             <p
-              className={`mx-auto md:m-0 text-3xl sm:text-5xl md:text-6xl pb-2 font-bold z-10 max-w-392px ${
```
chenjiahan, line 37: @sanyuan0704 It seems that `max-w-392px` is not a valid Tailwind CSS class name, should we delete it?

sanyuan0704: yeah

chenjiahan: done

sanyuan0704 (approved):

Issue #275, opened 2023-10-27
[Feature]: support configure preview plugin's default behavior


kainstar:
### What problem does this feature solve?

The preview plugin now treats all `jsx` or `tsx` code blocks as renderable React components, except for those with the `pure` meta. However, there are cases where we display the most of code blocks and render a small portion of them as a component.

Could we add an option property to control the default behavior of rendering code blocks as previews or simply displaying them?

### What does the proposed API look like?

1. add new option property in `@rspress/plugin-preview`

```ts
export type Options = {
  defaultBehavior: 'preview' | 'only-code'
}
```

2. add new meta identifier to enable preview mode for a code block
~~~mdx
```tsx preview
function App() {
  return "Hello World";
}
export default App;
```
~~~

sanyuan0704:
Good advice! Are you interested in implementing this feature?

kainstar:
Sure, I'll give it a shot later. Any suggestions on the API or property?

sanyuan0704:
It's better to keep the meta string same as the default behavior string.

PR #289, opened 2023-10-30
feat: support defaultInternalBehavior in preview and playground plugin

kainstar:
## Summary

Support `defaultInternalBehavior` in preview and playground plugin.

## Related Issue

https://github.com/web-infra-dev/rspress/issues/275

## By the way

With this PR user can use preview and playground plugin both for internal component. And for better DX, I suggest use `preview` and `playground` tag instead of current `code`, then user can manullay control how external component should be transformer, and `code` tag is changed to a simple way to render as normal code block.

If you agree with me and want more discussions, we can create an other new issue or discussion.

## Checklist


sanyuan0704:
Why the name is `defaultInternalBehavior` instead of `defaultBehavior`? I think the latter is easier.

packages/document/docs/en/plugin/official-plugins/playground.mdx
```diff
@@ -82,12 +82,21 @@ export default App;
 ```
 ````
 
+If you have set `defaultInternalBehavior` to `'pure'`, Rspress will not render this component by default, but instead display it as a code block. If you want to render a code block as a component in this situation, you can do add the `preview` identifier to specify, the usage is as follows:
+
+````md
```
sanyuan0704, line 87: The language should be `mdx`

kainstar:
> Why the name is `defaultInternalBehavior` instead of `defaultBehavior`? I think the latter is easier.

Because this option only decide how internal components render, if use `defaultBehavior`, it may give others the impression that it affects both internal and external components. If you perfer a shorter name, perhaps `internalBehavior` is also an option?

packages/document/docs/en/plugin/official-plugins/playground.mdx
kainstar: fixed

sanyuan0704:
On the one hand, the default behavior can be both applied in internal and external components, which is able to be implemented. For <code /> tag, we can parse the prop `renderMode`, which may be `pure` or `preview`.

On the other hand, the param name can be `defaultRenderMode`. i think this name is better.

sanyuan0704:
cc @kainstar

kainstar:
I have renamed `defaultInternalBehavior` to `defaultRenderMode`

sanyuan0704 (approved):

Issue #308, opened 2023-11-01
[Bug]:  Error wrapping in CodeBlock when set language = toml


recallwei:
### Version

```sh
System:
    OS: macOS 14.0
    CPU: (10) arm64 Apple M2 Pro
    Memory: 115.23 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 118.0.5993.117
    Safari: 17.0
  npmPackages:
    rspress: ^1.3.2 => 1.3.2
```

### Details

Expected to be:
<img width="1218" alt="image" src="https://github.com/web-infra-dev/rspress/assets/62941121/3f569b63-f987-4c48-a864-badb02b50cdf">
But:
<img width="996" alt="image" src="https://github.com/web-infra-dev/rspress/assets/62941121/7ae853ff-4806-4b06-8a7c-b270ef2806f8">

### Reproduce link

https://dolphin-admin-docs.bit-ocean.studio/guide/vue/deploy.html#%E4%BD%BF%E7%94%A8-netlify-%E6%89%98%E7%AE%A1

### Reproduce Steps

Use code block below:

````
```toml
[build.environment]
  NODE_VERSION = "20"

[build]
  publish = "dist"
  command = "pnpm build"

[[redirects]]
  from = "/base-api/*"
  to = "https://localhost:3000/:splat"
  status = 200
  force = true

[[redirects]]
  from = "/*"
  to = "/index.html"
  status = 200

[[headers]]
  for = "/manifest.webmanifest"
  [headers.values]
  Content-Type = "application/manifest+json"
```
````

PR #309, opened 2023-11-01
fix: error wrapping in CodeBlock when set language to toml

recallwei:
## Related Issue
#308 

## Checklist


recallwei:
I don't know if this is the right way to fix this bug.

image

image

---

After removing the `table` class, it looks ok:

image

Also does not affect the style of other CodeBlocks.

sanyuan0704:
LGTM

sanyuan0704 (approved):

PR #366, opened 2023-11-16
chore: simplify buildConfig

10Derozan:
## Summary
I think we can simpify the build config in repo, remove some unuse line to understand easily

## Checklist

### DIFF
--- a/packages/core/modern.config.ts
+++ b/packages/core/modern.config.ts
@@ -5,21 +5,6 @@ import { tailwindConfig } from './tailwind.config';
 const require = createRequire(import.meta.url);
 const tailwindPlugin = require('@modern-js/plugin-tailwindcss').default;
 
-const COMMON_EXTERNALS = [
-  'virtual-routes-ssr',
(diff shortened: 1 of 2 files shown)


packages/core/modern.config.ts
```diff
@@ -5,21 +5,6 @@ import { tailwindConfig } from './tailwind.config';
 const require = createRequire(import.meta.url);
 const tailwindPlugin = require('@modern-js/plugin-tailwindcss').default;
 
-const COMMON_EXTERNALS = [
```
chenjiahan, line 8: No need to external these packages?

10Derozan: 'runtime' is bundleless, so this can be moved, in fact, it is only used in `@rspress/theme-default`.
I guess these lines are migrated from modern doc, because 'theme-default' was inside 'core' before.

sanyuan0704:
LGTM

sanyuan0704 (approved):

PR #411, opened 2023-11-29
fix: remove dark mode script during build when disabled in config

EmNudge:
## Summary

Fixes the final build ignoring the user config setting `themeConfig.darkMode: false` (whereas it was previously respected in dev mode).

## Related Issue

https://github.com/web-infra-dev/rspress/issues/372


sanyuan0704:
Thanks

sanyuan0704 (approved):

PR #413, opened 2023-11-29
fix(core): support custom id when unable mdxRs

10Derozan:
## Summary
I found a limitation in https://github.com/web-infra-dev/rspress/blob/main/packages/document/docs/zh/guide/basic/use-mdx.mdx#L246.
I want to keep it same as mdxRs, so I draft this change.

## Checklist


.changeset/short-coins-add.md
```diff
@@ -0,0 +1,5 @@
+---
+'@rspress/core': patch
+---
+
+fix(core): support custom id when unable mdxRs
```
chenjiahan, line 5: unable -> disable 😄

packages/core/src/node/mdx/remarkPlugins/toc.ts
```diff
@@ -37,21 +38,23 @@ export const parseToc = (tree: Root) => {
...
 
-    // Collect h2 ~ h5
+    // Collect h2 ~ h4
     if (node.depth > 1 && node.depth < 5) {
-      const originText = node.children
+      let customId = '';
+      const text = node.children
         .map((child: ChildNode) => {
           if (child.type === 'link') {
             return child.children?.map(item => item.value).join('');
```
10Derozan, line 47: maybe we should extract text and id here too.

sanyuan0704 (approved):

PR #430, opened 2023-12-05
feat(theme-default): should open new tab/window on clicking with modifier keys

xc2:
## Summary

when clicking an internal link with modifier keys like `option/alt`, rspress should open an new tab as how the browser does instead of pushing history.

## Test Plan

- [x] `e2e/tests/link.test.ts`

## Checklist

- [x] I have updated the documentation. - no need


sanyuan0704:
The e2e test case in windows failed.

xc2:
> The e2e test case in windows failed.

fixed, would you please approve the workflow again?

sanyuan0704 (approved):

PR #435, opened 2023-12-06
fix: inline image data url start with slash prefix.

sumy7:
## Summary

When building with `pnpm run build`, ssg incorrectly adds a slash at the beginning of inline image resources in the generated HTML.

image

This pull request addresses this issue by checking if the URL is a data URL. If it is, the URL will be returned as is, without adding a slash.

This should ensure that inline image resources are correctly referenced in the generated HTML. Please review and provide any feedback.

## Checklist


sanyuan0704 (approved):

sanyuan0704:
Thank you

Issue #453, opened 2023-12-13
[Feature]: Improve error logs if missing locales config


chenjiahan:
### What problem does this feature solve?

If user configures `lang`, but forget to add `locales`, the `rspress dev` command will fail:

<img width="1197" alt="Screenshot 2023-12-13 at 14 14 19" src="https://github.com/web-infra-dev/rspress/assets/7237365/a9ca2a09-88a0-4010-9bf2-151d9249221c">

This error log is not very clear, we need to provide more clear error information.

You can reproduce this error in https://github.com/chenjiahan/rspress-repro-missing-locale

1. pnpm i
2. pnpm dev

### What does the proposed API look like?

Improve error logs, tell user to configure locales correctly.

PR #454, opened 2023-12-13
feat: improve errors outputted to console

GODLiangCY:
## Summary
The error in #453 was caused by the `@rspress/plugin-auto-nav-sidebar`, which can't generate configs correctly if `locales` are not configured. [Promise here](https://github.com/web-infra-dev/rspress/blob/main/packages/plugin-auto-nav-sidebar/src/walk.ts#L56-L118) is rejected because the value of the name field is `undefined`
image

Not sure if we should add this to the suitable part of the documentation and where the "suitable part" is

## Related Issue

#453 

## Checklist

### DIFF
--- a/.changeset/slimy-deers-prove.md
+++ b/.changeset/slimy-deers-prove.md
@@ -1,0 +1,5 @@
+---
+'@rspress/plugin-auto-nav-sidebar': patch
+---
+
+feat: improve error logs if missing locales config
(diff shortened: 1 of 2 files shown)


sanyuan0704:
Thank you.

sanyuan0704 (approved):

chenjiahan:
👍👍

PR #489, opened 2023-12-26
chore: degrade rspack to 0.4.0

sanyuan0704:
## Summary

The build performance has been declined from 0.4.1. So we fallback to 0.4.0 to ensure the origin build performance. 

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,6 +49,11 @@
       "packages/*"
     ]
   },
+  "pnpm": {
+    "overrides": {
+      "@rspack/core": "0.4.0"
+    }
+  },
   "devDependencies": {
     "@manypkg/get-packages": "^2.2.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -49,6 +49,11 @@
       "packages/*"
     ]
   },
+  "pnpm": {
+    "overrides": {
+      "@rspack/core": "0.4.0"
```
chenjiahan, line 54: Maybe we should downgrade Rsbuild version? Pnpm overrides can not work for Rspress users.

sanyuan0704: That means we should revert the following pr:
- https://github.com/web-infra-dev/rspress/pull/440
- https://github.com/web-infra-dev/rspress/pull/439

chenjiahan: yeah...It may be more better to push Rspack to fix the performance issue

PR #543, opened 2024-01-12
chore(deps): bump rsbuild to 0.3.1

9aoy:
## Summary

bump rsbuild from 0.2.1 to 0.3.1.

 there are some breaking change in rsbuild 0.3.0: https://github.com/web-infra-dev/rsbuild/releases/tag/v0.3.0

## Checklist


packages/core/src/node/dev.ts
```diff
@@ -18,14 +18,16 @@ interface DevOptions {
...
-  const base = config?.base ?? '';
   const isProd = false;
   const pluginDriver = new PluginDriver(config, isProd);
   await pluginDriver.init();
 
   try {
     const modifiedConfig = await pluginDriver.modifyConfig();
     await pluginDriver.beforeBuild();
+
+    rsbuildLogger.override(logger);
```
chenjiahan, line 29: Both Rsbuild and Rspress uses `rslog`, I think the override is no longer needed.

chenjiahan (approved):

PR #547, opened 2024-01-14
refactor: use rsbuild.preview for previewing

chenjiahan:
## Summary

Use `rsbuild.preview` for previewing, which means:

- less dependencies, removing polka and compression.
- align the behavior with `rspress dev`.
- `builderConfig` can also work when running preview.

image

## Related Issue

https://rsbuild.dev/api/javascript-api/instance#rsbuildpreview

## Checklist


packages/core/src/node/serve.ts
```diff
@@ -22,35 +14,31 @@ export async function serve(options: ServeOptions) {
...
+  const modifiedConfig = await pluginDriver.modifyConfig();
+
+  const builder = await initRsbuild(
+    config.root,
+    modifiedConfig,
+    pluginDriver,
+    false,
+  );
+
+  await builder.preview();
```
sanyuan0704, line 43: Could the server take the `base` path in preview log?

chenjiahan: Yes, `base` is supported in `createInternalBuildConfig`, we do not need to add extra code here 😄

image

sanyuan0704: Awesome

sanyuan0704 (approved):

PR #549, opened 2024-01-15
feat: introduce biome to rspress

zoolsher:
## Summary

use biome as formatter

## Related Issue

// SKIP

## Checklist

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -61,7 +61,7 @@
   "editor.codeActionsOnSave": {
     "source.fixAll.eslint": "explicit"
   },
-  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.defaultFormatter": "biomejs.biome",
   "javascript.format.enable": false,
   "typescript.format.enable": false,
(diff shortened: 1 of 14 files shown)


.vscode/settings.json
```diff
@@ -76,7 +76,8 @@
     "editor.tabSize": 2,
     "editor.formatOnType": true,
     "editor.formatOnPaste": true,
-    "editor.formatOnSave": true
+    "editor.formatOnSave": true,
+    "editor.defaultFormatter": "biomejs.biome"
```
chenjiahan, line 80: I have also tried using Biome as the formatter, but Biome does not yet support formatting `.md` and `.mdx`files, so we may still need to preserve prettier to format unsupported files.

chenjiahan: see https://biomejs.dev/internals/language-support/

sanyuan0704: We can configure the formatter of md(x) files as prettier

sanyuan0704 (approved):

PR #554, opened 2024-01-17
feat: reuse logger from Rsbuild

chenjiahan:
## Summary

Reuse logger from Rsbuild. Rsbuild now provides logger as a public API see: https://rsbuild.dev/api/javascript-api/core#logger.

This change can also upgrade rslog to latest version (v1.2.0) and fix some bug, see https://github.com/rspack-contrib/rslog/releases

## Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
 
 settings:
   autoInstallPeers: true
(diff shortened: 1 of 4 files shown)


pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.1'
```
SoonIter, line 1: Is it the installation result with pnpm@8.6.1 ?

chenjiahan: fixed, the corepack not working..

SoonIter (approved):

Issue #133, opened 2023-10-01
Rspress Roadmap


sanyuan0704:
### What problem does this feature solve?

xx

### What does the proposed API look like?

xx

jeffwcx:
Are there any plans to add component display features, similar to storybooks?

sanyuan0704:
> Are there any plans to add component display features, similar to storybooks?

Yeah, Rspress has official preview and playground plugin for component development scene.

https://rspress.dev/plugin/official-plugins/preview
https://rspress.dev/plugin/official-plugins/playground

And if you have any need else,  welcome to feedback,

jeffwcx:
I have installed and experienced the writing of the demo. It is very fast and the experience is very good, but there are still some points that can be enhanced.
1. Lack of CodeSandbox, Stackblitz, CodePen support.
2. Only supports react component. It is recommended to support cross-framework component display capabilities, such as Vue components.
3. The ability to extract information from react/vue components and generate API tables

sanyuan0704:
> I have installed and experienced the writing of the demo. It is very fast and the experience is very good, but there are still some points that can be enhanced.
> 
> 1. Lack of CodeSandbox, Stackblitz, CodePen support.
> 2. Only supports react component. It is recommended to support cross-framework component display capabilities, such as Vue components.
> 3. The ability to extract information from react/vue components and generate API tables

Good advice! I will response to the three points:
1. The implement of component display in Rspress does not depend on any external sandbox such as `CodeSandbox`、`Stackblitz`, etc. But we will also support the link to jump to the corresponding platform in the future. cc @10Derozan 
2. The custom component render process is based on React as well as components in the theme.So we have no plan to support vue component in preview/playground plugin.
3. The ability is under construction and we have integrated `react-docgen-typescript` and `documentation` to generate API Table automatically.Stay tuned!

jeffwcx:
For the second point, maybe you can run instances of other frameworks in the react framework. Storybook also uses this pattern. Maybe I can make a PR later;
For the third point, it would be better to have an interface to connect to other frameworks.

zoolsher:
@jeffwcx Having cross-framework support as a feature in the framework would be fantastic! looking forward to your PR

Yukiniro:
Hey guys, what should I do if I want to contribute something such as the badge component for the roadmap?

sanyuan0704:
#542

sanyuan0704:
[...]

PR #563, opened 2024-01-20
feat: framework agnostic component preview solution

sanyuan0704:
## Summary

Preview component demo for other framework instead of React-only.

## Related Issue

#133 

## Checklist


10Derozan:
Maybe we need a boolean configure to distinguish react and solidjs, because their entry is not the same.

10Derozan:
This has been replaced by #578

PR #564, opened 2024-01-20
fix: ensure afterBuild executed after dev compile done

sanyuan0704:
## Summary

In rsbuild, when we called `builder.startDevServer`, the compile process doesn't finished. So we should execute `afterBuild` hook of Rspress in the `onDevCompileDone` hook of Rsbuild to ensure the compile process done.
## Checklist

### DIFF
--- a/packages/core/src/node/dev.ts
+++ b/packages/core/src/node/dev.ts
@@ -34,12 +34,21 @@ export async function dev(options: DevOptions): Promise<ServerInstance> {
       false,
       extraBuilderConfig,
     );
+    builder.addPlugins([
+      {
+        name: 'rspress:afterBuild',
+        setup(api) {
+          api.onDevCompileDone(async () => {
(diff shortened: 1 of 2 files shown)


packages/core/src/node/dev.ts
```diff
@@ -34,12 +34,21 @@ export async function dev(options: DevOptions): Promise<ServerInstance> {
       false,
       extraBuilderConfig,
     );
+    builder.addPlugins([
```
chenjiahan, line 37: We can also use `rsbuild.onDevCompileDone`

https://rsbuild.dev/api/javascript-api/instance#rsbuildondevcompiledone

sanyuan0704: great

sanyuan0704: The API of Rsbuild is so smooth!

chenjiahan (approved):

PR #566, opened 2024-01-20
fix: ensure frontmatter exists during parsing mdx content

### LINKED ISSUE #534: [Bug]: "frontmatter is not defined" error in MDX
### Version

```sh
System:
  OS: Windows 11 10.0.22621
  CPU: (16) x64 AMD Ryzen 7 7800X3D 8-Core Processor
  Memory: 16.83 GB / 31.15 GB
Browsers:
  Edge: Chromium (120.0.2210.121)
  Internet Explorer: 11.0.22621.1
npmPackages:
  rspress: ^1.9.3 => 1.9.3
```

### Details

I followed the example in [Docs](https://rspress.dev/guide/basic/use-mdx.html#front-matter), and it causes an error.

`docs/hello.mdx`
```mdx
---
title: Hello World
---

# {frontmatter.title}
```

```
hello.mdx:7 Uncaught ReferenceError: frontmatter is not defined
    at _createMdxContent (hello.mdx:7:1)
```

### Reproduce link

[Codesandbox](https://codesandbox.io/p/devbox/nifty-tristan-hx99f4?file=%2Fdocs%2Fguide%2Fhello.mdx%3A5%2C22)

### Reproduce Steps

1. Click 'Guide' in navigation bar
2. Click '{frontmatter.title}' in side bar
3. See the blank white screen, you can see the error in console

GODLiangCY:
## Summary

TODO: write here til pr finished

## Related Issue

closes #534 

## Checklist

### DIFF
--- a/packages/plugin-auto-nav-sidebar/src/utils.ts
+++ b/packages/plugin-auto-nav-sidebar/src/utils.ts
@@ -28,16 +28,21 @@ export async function detectFilePath(rawPath: string) {
 export async function extractH1Title(
   filePath: string,
   rootDir: string,
+  /** prefer using the title declared in frontmatter rather than file content */
+  preferFrontMatter = true,
 ): Promise<string> {
   try {
     const realPath = await detectFilePath(filePath);
     const content = await fs.readFile(realPath, 'utf-8');
+    const { frontmatter } = loadFrontMatter(content, filePath, rootDir);
+    if (frontmatter.title && preferFrontMatter) {
+      return frontmatter.title;
+    }
     const fileNameWithoutExt = path.basename(realPath, path.extname(realPath));
     const h1RegExp = /^#\s+(.*)$/m;
     const match = content.match(h1RegExp);
     if (!match) {
-      const { frontmatter } = loadFrontMatter(content, filePath, rootDir);
-      return frontmatter.title || fileNameWithoutExt;
+      return fileNameWithoutExt;
     } else {
       return match[1] || fileNameWithoutExt;
     }


GODLiangCY:
Well, the logic seems somewhat confusing to me😂. The [official doc](https://rspress.dev/guide/basic/use-mdx.html#front-matter) of Rspack says 
> "The previously defined properties will be passed to the component as frontmatter properties. "

Thus I guess this should be processed by the `compile` function imported from `@rspress/mdx-rs` in `packages/core/src/node/runtimeModule/siteData/extractPageData.ts`, but the compiled result didn't take frontmatter as a property.
image


Am I missing something, or should I make a PR to `@rspress/mdx-rs` first? @sanyuan0704 sorry for any interruption by ping directly, but I desperately need your advice.

P.S: If a new PR to `@rspress/mdx-rs` is needed, I'm willing to give it a try, but it may be very slow because I'm a Rust newbie

packages/plugin-auto-nav-sidebar/src/utils.ts
```diff
@@ -28,16 +28,21 @@ export async function detectFilePath(rawPath: string) {
 export async function extractH1Title(
   filePath: string,
   rootDir: string,
+  /** prefer using the title declared in frontmatter rather than file content */
+  preferFrontMatter = true,
```
sanyuan0704, line 32: Why we perfer using `frontmatter.title`? For performance, we used the  h1 title in origin content and fallback to the frontmatter.title when  h1 title does not exist.

sanyuan0704:
> Well, the logic seems somewhat confusing to me😂. The [official doc](https://rspress.dev/guide/basic/use-mdx.html#front-matter) of Rspack says
> 
> > "The previously defined properties will be passed to the component as frontmatter properties. "
> 
> Thus I guess this should be processed by the `compile` function imported from `@rspress/mdx-rs` in `packages/core/src/node/runtimeModule/siteData/extractPageData.ts`, but the compiled result didn't take frontmatter as a property. image
> 
> Am I missing something, or should I make a PR to `@rspress/mdx-rs` first? @sanyuan0704 sorry for any interruption by ping directly, but I desperately need your advice.
> 
> P.S: If a new PR to `@rspress/mdx-rs` is needed, I'm willing to give it a try, but it may be very slow because I'm a Rust newbie

In my opinion, there is no need to change `@mdx-rs` package. It's a bug of the mdx loader in rspress https://github.com/web-infra-dev/rspress/blob/main/packages/core/src/node/mdx/loader.ts. We should inject the `const frontmatter = xxx` statement in the loader compile result.

sanyuan0704:
Reference: https://github.com/web-infra-dev/rspress/pull/569

packages/plugin-auto-nav-sidebar/src/utils.ts
GODLiangCY: If using h1 title in the origin content, then in #534 's case, the sidebar will look like
image

That may not meet the developer's need

sanyuan0704: All right.

sanyuan0704: Maybe we can create a front matter cache manager in `@rspress/shared` to parse and reuse all the frontmatter info  of md(x) files to avoid duplicated parse process.

GODLiangCY: > If using h1 title in the origin content, then in #534 's case, the sidebar will look like image
> 
> That may not meet the developer's need

Or this case shows a wrong usage? If so, we should not prefer using `frontmatter.title`

sanyuan0704: yeah, i think it shows a unexpected usage.

PR #578, opened 2024-01-25
feat(plugin-preview): support to preview other framework not only react

10Derozan:
## Summary
Try to run a new build instance to support demo.

## Checklist


packages/plugin-preview/src/index.ts
```diff
@@ -60,204 +60,102 @@ export const demoMeta: Record<
...
-          .join(',')}];
-        `;
-      demoRuntimeModule.writeModule('virtual-meta', virtualMeta);
+    },
+    beforeBuild(_, isProduction) {
+      isProd = isProduction;
+    },
+    async afterBuild(config, isProd) {
+      if (isMobile && iframePosition === 'fixed') {
+        // TODO: write entry for all page
```
10Derozan, line 94: We need write a entry when show device, need preview all demo in one route. like [virtual-demo](https://github.com/web-infra-dev/rspress/blob/main/packages/plugin-preview/src/virtual-demo.tsx#L20-L29)

10Derozan: Unfortunately, this cannot be handled at runtime.
We must concatenate strings and manually output them as files.

10Derozan:
Close second devServer after rsbuild release https://github.com/web-infra-dev/rsbuild/pull/1504, because rspress cli may restart when config file is changed.

PR #581, opened 2024-01-25
fix(plugin-preview): Convert isMobileMode variable to boolean

mattuylee:
When use html code element with isMobile prop, it renders wrongly because the `isMobileMode` variable should be boolean but is an object. This commit forcibly convert `isMobileMode` to boolean type.

## Checklist

### DIFF
--- a/.changeset/violet-falcons-listen.md
+++ b/.changeset/violet-falcons-listen.md
@@ -1,0 +1,5 @@
+---
+'@rspress/plugin-preview': patch
+---
+
+Support external demo to specific isMobile prop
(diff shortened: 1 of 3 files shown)


10Derozan:
Can you take a example? I don't know the bad case.

mattuylee:
> Can you take a example? I don't know the bad case.

like this:

```mdx
<code src="@/stories/h5/selector/basic" isMobile={true} />
```
Mobile mode doesn't work without this commit change.

10Derozan:
Thanks, you are right, but I fount another problem:
when we use isMobile props in code element,  the node value is a `mdxJsxAttributeValueExpression`,
so we can not  use `isMobile={false}` though you convert it to Boolean.

10Derozan (approved):

PR #605, opened 2024-02-07
fix(plugin-auto-nav-sidebar): use logical OR instead of nullish coalescing when label is empty

10Derozan:
## Summary
When use `type: 'file'` in _meta.json, can not extract h1 title 

## Checklist


10Derozan:
use default value, so nullish coalescing is not correct.
https://github.com/web-infra-dev/rspress/pull/605/files#diff-74615b4d58f7bf064817b05a66382330230ce484a2c79288530b1635be640436L75

PR #671, opened 2024-02-26
fix: memoize the callback in `useI18n` hook

ambar:
## Summary

This makes it possible to use in hook dependencies, works with ESLint, and to avoid unnecessary re-render.

## Checklist


sanyuan0704 (approved):

sanyuan0704:
You can run `pnpm run change` to add a changeset of patch type.

sanyuan0704 (approved):

PR #678, opened 2024-02-27
fix(theme-default): scroll search panel when current item is unvisible

10Derozan:
## Checklist


10Derozan:
What choice we use?
1. when keydown or keyup, we scroll the search panel.
2. Only when the selected item is unvisible, we scroll the panel to see selected item.

10Derozan:
> What choice we use?
> 
> 1. when keydown or keyup, we scroll the search panel.
> 2. Only when the selected item is unvisible, we scroll the panel to see selected item.

@chenjiahan
Example
image
What choice we use?

chenjiahan:
What is the context of this feature, why do we need to scroll the search panel?

10Derozan:
> What is the context of this feature, why do we need to scroll the search panel?

You can see the picture above your comment.
Our panel support to choose the item by keydown. And when the search result is much than visible window, we need to scroll to let user see the item.

chenjiahan:
ok, we can listen to keydown and scroll to the selected option, just like the algolia search component (see https://vuejs.org/)

10Derozan:
> ok, we can listen to keydown and scroll to the selected option, just like the algolia search component (see https://vuejs.org/)

It exist problem too. 
image

chenjiahan:
we can also listen to the input event if it is required.

10Derozan:
Now I can only choose the planA, scroll after keydown, because I can not get the all item height.

PR #680, opened 2024-02-28
fix(core): remove global to avoid ambiguity in non-greedy match in flat mdx

10Derozan:
## Summary
Fix 
image

## Checklist


sanyuan0704:
I still cannot  understand why appearing this error:

image

Even if the regex expression match failed, it could not influence the resolve process.

10Derozan:
I argee with you, maybe you can debug it., But I succed when I change the match rule.

PR #677, opened 2024-02-27
fix(theme-default): doc container with full width.

sumy7:
## Summary

this pr use `width: 100%` not `calc(100vw - sidebar_width - outline_width)` to make doc container fill the entire width while screen width `>= 1280px` and `< 1440px`.

## Related Issue

https://github.com/web-infra-dev/rspress/pull/643#issuecomment-1964202075

## Checklist

- [ ] I have added changeset via `pnpm run change`.
- [ ] I have updated the documentation.
- [ ] I have added tests to cover my changes.

PR #693, opened 2024-03-01
fix: make outline have a well width

sumy7:
## Summary

This PR adds style `overflow-hidden` to the doc container to ensure the code block scrolls correctly when there is no wrap.

## Related Issue

https://github.com/web-infra-dev/rspress/pull/677#issuecomment-1971280511

## Checklist

### DIFF
--- a/packages/theme-default/src/layout/DocLayout/index.tsx
+++ b/packages/theme-default/src/layout/DocLayout/index.tsx
@@ -74,7 +74,7 @@ export function DocLayout(props: DocLayoutProps) {
       <div
         className={`${styles.content} rspress-doc-container flex flex-shrink-0 mx-auto`}
       >
-        <div className="w-full">
+        <div className="w-full overflow-hidden">
           {isOverviewPage ? (
             <Overview content={docContent} />
           ) : (
--- a/packages/theme-default/src/layout/DocLayout/index.tsx
+++ b/packages/theme-default/src/layout/DocLayout/index.tsx
@@ -74,7 +74,7 @@ export function DocLayout(props: DocLayoutProps) {
       <div
         className={`${styles.content} rspress-doc-container flex flex-shrink-0 mx-auto`}
       >
-        <div className="w-full overflow-hidden">
+        <div className="w-full overflow-hidden flex-1">
           {isOverviewPage ? (
             <Overview content={docContent} />
           ) : (
--- a/.changeset/shy-eagles-laugh.md
+++ b/.changeset/shy-eagles-laugh.md
@@ -1,0 +1,5 @@
+---
+"@rspress/theme-default": patch
+---
+
+fix(theme-default): make outline have a well width again
--- a/packages/theme-default/src/layout/DocLayout/index.module.scss
+++ b/packages/theme-default/src/layout/DocLayout/index.module.scss
@@ -104,9 +104,11 @@
     padding: 48px 0 72px 0;
     :global(.rspress-doc),
     :global(.rspress-doc-footer) {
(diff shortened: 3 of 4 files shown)


sumy7:
I've looked into the issue and it appears that the code block, when not wrapped to a new line, is expanding the container, which in turn is shrink the width of the outline. I've added `overflow-hidden` to allow the container to properly constrain the width of the code block.

chenjiahan:
In addition to the code block, could other elements also possibly expand the container?

Setting a fixed container width may solve this problem better.

sanyuan0704:
> While using a fixed width could solve the issue, the current problem is that the declared width of the sidebar (368px) does not match its actual width (~280px). This discrepancy leads to errors when calculating the fixed width using CSS variables. Therefore, I would still prefer to use a responsive layout approach to address this issue. Or Should we reconsider/redesign the responsive rules for the whole layout?

Thanks for reminding me the fact. I also prefer to use a responsive layout approach like the original way

sanyuan0704 (approved):

packages/theme-default/src/layout/DocLayout/index.tsx
```diff
@@ -74,7 +74,7 @@ export function DocLayout(props: DocLayoutProps) {
       <div
         className={`${styles.content} rspress-doc-container flex flex-shrink-0 mx-auto`}
       >
-        <div className="w-full">
+        <div className="w-full overflow-hidden flex-1">
```
chenjiahan, line 77: @sumy7 can we remove the `overflow-hidden` or add a new class-name to this element?

The Rsbuild document is broken by the `overflow-hidden` style, as Rsbuild uses a custom overview component in this page: https://deploy-preview-1713--rsbuild.netlify.app/config/index

Screenshot 2024-03-03 at 21 51 30

sumy7: @chenjiahan removing overflow-hidden sounds like a good idea.
@sanyuan0704 , could you help fix this ? Or I will  submit a new PR tomorrow.

Also, for this page, I suggest disabling the display of the outline in the frontmatter to achieve a full-screen effect.

sanyuan0704: The `overflow-hidden` style seems redundant. Can you explain the influence after removing it? Perhaps there are some cases i didn't notice. And i'm not convenient now, expect your new PR! ❤️

sumy7: Hi @sanyuan0704.
Let me explain the purpose of `overflow-hidden` here.
When I initially added `overflow-hidden`, the intention was to use the BFC to prevent long, non-wrapping sentences in child elements from exceeding the restricted width and compressing the outline.
However, following @chenjiahan 's suggestion, I switched to using a fixed width and added `flex-1` to ensure the space occupied by the outline.
As a result, the original purpose of `overflow-hidden` to limit the width of long sentences was lost.

This style layout modification has brought up many cases that I hadn't considered, as reported by @chenjiahan from rsbuild. However, the approach of fixing problems as they arise can easily shake my confidence. I'm considering how to avoid situations where fixing one issue leads to more problems.

chenjiahan: Thanks for your detailed explanation!

As you said, it is very easy to introduce some small breaking changes or edge cases when modifying CSS styles. It is also hard to testing it.

From my perspective, we can do :

1. Stabilize the main CSS styles and avoiding changing them frequently.
2. Add more class names to main elements to allow users overriding the styles.

PR #643, opened 2024-02-20
fix: fix some theme style issues.

sumy7:
## Summary

This pr fix theme style issues when navbar or outline not display.

1. change css variable `--rp-aside-width` and `--rp-sidebar-width` from `0` to `0px` 
image
while chrome tract `calc(100% - 0px)` as invalid property value

2. move sidebar padding property to a sub container
image
while sidebar width set to 0 still has padding-right remind.

3. `.rspress-doc` and `.rspress-doc-footer` change `box-sizing: content-box` to `box-sizing:border-box` in screen 1280px because of `content-width + padding > 100vw`

4. increase outline `max-height` for reduce the gap between outline bottom and page bottom.
image

## Checklist

- [ ] I have added changeset via `pnpm run change`.
- [ ] I have updated the documentation.
- [ ] I have added tests to cover my changes.

PR #701, opened 2024-03-05
fix: remove useless overflow style, which cause api page width collapsed

sanyuan0704:
## Summary

The `overflow-hidden` style cause the api page width collapsed. The preview after removing it:

image

In the meantime, the expanding problem of doc content has been solved by #693 through setting an appropriate outline width:

image

## Related Issue

#643 

## Checklist

### DIFF
--- a/packages/theme-default/src/layout/DocLayout/index.tsx
+++ b/packages/theme-default/src/layout/DocLayout/index.tsx
@@ -74,7 +74,7 @@ export function DocLayout(props: DocLayoutProps) {
       <div
         className={`${styles.content} rspress-doc-container flex flex-shrink-0 mx-auto`}
       >
-        <div className="w-full">
+        <div className="w-full flex-1">
           {isOverviewPage ? (
             <Overview content={docContent} />
           ) : (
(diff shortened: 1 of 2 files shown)


sanyuan0704:
cc @chenjiahan @sumy7

packages/theme-default/src/layout/DocLayout/index.tsx
```diff
@@ -74,7 +74,7 @@ export function DocLayout(props: DocLayoutProps) {
       <div
         className={`${styles.content} rspress-doc-container flex flex-shrink-0 mx-auto`}
       >
-        <div className="w-full overflow-hidden flex-1">
+        <div className="w-full">
```
chenjiahan, line 77: Can we remove `flex-1`? It seems still useful:

image

sumy7: We cannot remove the `flex-1` here, this property is crucial to ensure that the width of the outline always remains at 268px. Without this property, the widths of the container and outline would be redistributed due to flex.

sanyuan0704: I get it

sumy7:
We should add a property to prevent the outline's width from shrinking, such as `flex-shrink-0` at outline or the previously used `flex-1` at doc container.

the outline width before use `flex-1`
image
after
image

sanyuan0704:
> `flex-shrink-0`

Yeah, i use the `flex-1` in doc content to reserve the outline width.

sumy7 (approved):

chenjiahan (approved):

PR #702, opened 2024-03-05
chore(workflow): use Biome as the linter

chenjiahan:
## Summary

Use Biome as the linter to replace ESLint, reduced dependencies and improved lint performance.

image

image

## TODO

- enable more rules: I have disabled many lint rules in `biome.json`, we should enable these rules and fix the reported issues later.
- lint more files: currently only `packages/**/*.ts` are linted, we can lint more files later.

## Checklist


sanyuan0704:
Wow

10Derozan:
Fast!

PR #716, opened 2024-03-06
perf: remove the builtin @rsbuild/plugin-svgr

chenjiahan:
## Summary

Remove the built-in SVGR plugin since Rspress does not rely on this plugin.

The SVGR feature is not mentioned in the documentation, but some users who have used custom React components may rely on this behavior, so this could be a potentially breaking change.

## Benefits

The benefits of removing the SVGR plugin are obvious, it can significantly reduce the number of dependencies (-80), and it also allows the use of different SVGR syntax or config, so we will remove it in the next minor version of Rspress.

Screenshot 2024-03-06 at 11 58 29

## Migration

Users who have SVGR needs can manually use the [@rsbuild/plugin-svgr](https://rsbuild.dev/plugins/list/plugin-svgr):

1. run `pnpm add @rsbuild/plugin-svgr -D`
2. register this plugin in `rspress.config.ts`:

```js
import { pluginSvgr } from '@rsbuild/plugin-svgr';

export default defineConfig({
  builderConfig: {
    plugins: [
      pluginSvgr({
        svgDefaultExport: 'component',
      }),
    ],
  }
});
```

## Checklist

### DIFF
--- a/.changeset/kind-insects-develop.md
+++ b/.changeset/kind-insects-develop.md
@@ -1,0 +1,5 @@
+---
+"@rspress/core": minor
+---
+
+perf: remove the builtin @rsbuild/plugin-svgr
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -55,7 +55,6 @@
     "@modern-js/utils": "2.48.0",
     "@rsbuild/core": "0.4.11",
     "@rsbuild/plugin-react": "0.4.11",
-    "@rsbuild/plugin-svgr": "0.4.11",
     "@rspress/theme-default": "workspace:*",
     "@rspress/runtime": "workspace:*",
     "@rspress/plugin-medium-zoom": "workspace:*",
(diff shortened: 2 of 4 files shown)


10Derozan (approved):

sanyuan0704:
I think the default theme will have svg render problem after executing `DOC_DEBUG=true pnpm dev`

sanyuan0704:
cc @10Derozan

10Derozan:
> > I think the default theme will have svg render problem after executing DOC_DEBUG=true pnpm dev
> 
> Can you provide more info for this?

Debug mode will use source code
https://github.com/web-infra-dev/rspress/blob/main/packages/theme-default/src/node/source-build-plugin.ts#L19

10Derozan:
Have you test IconWrapper which you suggested?

chenjiahan:
Ok, I get it.

I can not reproduce the problem because the SVG icons are always resolved to the `dist` folder now:

image

chenjiahan:
> Have you test IconWrapper which you suggested?

Implemented in https://github.com/web-infra-dev/rspress/pull/720.

It works and allows us to use both string SVG or component SVG at the same time.

sanyuan0704 (approved):

sanyuan0704:
> Ok, I get it.
> 
> I can not reproduce the problem because the SVG icons are always resolved to the `dist` folder now:
> 
> image

Get it

PR #723, opened 2024-03-06
fix(theme-default): correct lark svg icon

fi3ework:
## Summary

Only the lark icon specified the width and height, causing the demo to be distorted.

before:
image

after:
image

## Checklist


chenjiahan (approved):

chenjiahan:
Nice👍

PR #724, opened 2024-03-06
feat(core): support func type when change builderConfig in plugin

10Derozan:
## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -239,15 +239,13 @@ export async function initRsbuild(
     pluginDriver,
   );
 
+  const builderConfigList = [...pluginDriver.getPlugins().map(plugin => plugin.builderConfig ?? {}), config?.builderConfig || {}, extraRsbuildConfig || {}]
+  const rsbuildConfig = builderConfigList.reduce<RsbuildConfig>((acc, cur): RsbuildConfig => {
+    return cur instanceof Function ? cur(acc, isSSR) || acc : mergeRsbuildConfig(acc, cur);
+  }, internalRsbuildConfig)
+
   const rsbuild = await createRsbuild({
-    rsbuildConfig: mergeRsbuildConfig(
-      internalRsbuildConfig,
-      ...(pluginDriver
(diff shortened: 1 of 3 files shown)


packages/shared/src/types/Plugin.ts
```diff
@@ -33,7 +33,7 @@ export interface RspressPlugin {
   /**
    * Rsbuild config.
    */
-  builderConfig?: RsbuildConfig;
+  builderConfig?: RsbuildConfig | ((config: RsbuildConfig, isSSR: boolean) => void | RsbuildConfig);
```
chenjiahan, line 36: the second param should be an object, making it  easier to add more params in the future.

packages/core/src/node/initRsbuild.ts
```diff
@@ -239,15 +239,13 @@ export async function initRsbuild(
     pluginDriver,
   );
 
+  const builderConfigList = [...pluginDriver.getPlugins().map(plugin => plugin.builderConfig ?? {}), config?.builderConfig || {}, extraRsbuildConfig || {}]
+  const rsbuildConfig = builderConfigList.reduce<RsbuildConfig>((acc, cur): RsbuildConfig => {
+    return cur instanceof Function ? cur(acc, isSSR) || acc : mergeRsbuildConfig(acc, cur);
+  }, internalRsbuildConfig)
+
```
chenjiahan, line 246: can we add a test case for this change

chenjiahan: also document

packages/shared/src/types/Plugin.ts
chenjiahan: The "config" hook supports function type, maybe we should use it to modify builderConfig.

10Derozan: The config hook don't have internalRsbuildConfig, derived from the rspress configuration.
The config hook still process rspress config,.

chenjiahan: Sure, if you need to process the builtin Rsbuild config, you can write an Rsbuild plugin.

PR #744, opened 2024-03-08
feat(theme): support custom overview and sub overview

10Derozan:
## Summary
Add support for sub dir overview, we can filter the sidebar groups by link.

## Checklist

### DIFF
--- a/.changeset/orange-waves-admire.md
+++ b/.changeset/orange-waves-admire.md
@@ -1,0 +1,5 @@
+---
+'@rspress/theme-default': minor
+---
+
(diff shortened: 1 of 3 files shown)


packages/document/docs/zh/api/config.mdx
```diff
@@ -0,0 +1,3 @@
+---
+overview: true
```
10Derozan, line 2: Test case

10Derozan: image

PR #808, opened 2024-03-18
fix: custom theme color not work

### LINKED ISSUE #743: [Bug]: Custom CSS variables not work after bumping Rspress v1.14.0
### Version

```sh
System:
    OS: macOS 13.6.3
    CPU: (10) arm64 Apple M1 Pro
    Memory: 1.24 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 122.0.6261.111
    Safari: 16.6
  npmPackages:
    @rspress/shared: 1.14.0 => 1.14.0
    rspress: 1.14.0 => 1.14.0
```

### Details

Custom CSS variables not work after bumping Rspress v1.14.0.

- Rspack website with `rspress@1.11.0`:

image

- Rspack website with `rspress@1.14.0`:

image

It seems to be a problem with the CSS order. CSS variables in the [theme/index.css](https://github.com/web-infra-dev/rspack-website/blob/bump_rspress_0308/theme/index.css) can not override the default CSS variables.

### Reproduce link

https://github.com/web-infra-dev/rspack-website/tree/bump_rspress_0308

### Reproduce Steps

1. pnpm i
2. pnpm dev

sanyuan0704:
## Summary

Fix #743 

## Checklist

### DIFF
--- a/packages/core/src/runtime/App.tsx
+++ b/packages/core/src/runtime/App.tsx
@@ -10,10 +10,12 @@ import { HelmetProvider } from 'react-helmet-async';
 import React, { useContext, useLayoutEffect } from 'react';
 import { Header, PageData, cleanUrl, MDX_REGEXP } from '@rspress/shared';
 import globalComponents from 'virtual-global-components';
-import 'virtual-global-styles';
 
 // eslint-disable-next-line import/no-commonjs
 const { default: Theme } = require('@theme');
+
+// Inject the global styles come from plugins or config file
+require('virtual-global-styles');
 
 export enum QueryStatus {
   Show = '1',
@@ -39,7 +41,9 @@ export async function initPageData(routePath: string): Promise<PageData> {
     const extractPageInfo = siteData.pages.find(page => {
       const normalize = (p: string) =>
         // compat the path that has no / suffix and ignore case
-        p.replace(/\/$/, '').toLowerCase();
+        p
+          .replace(/\/$/, '')
+          .toLowerCase();
       return isEqualPath(normalize(page.routePath), normalize(routePath));
     });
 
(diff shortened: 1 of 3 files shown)


packages/core/src/runtime/App.tsx
```diff
@@ -10,11 +10,13 @@ import { HelmetProvider } from 'react-helmet-async';
 import React, { useContext, useLayoutEffect } from 'react';
 import { Header, PageData, cleanUrl, MDX_REGEXP } from '@rspress/shared';
 import globalComponents from 'virtual-global-components';
-import 'virtual-global-styles';
 
 // eslint-disable-next-line import/no-commonjs
 const { default: Theme } = require('@theme');
 
+// Inject the global styles come from plugins or config file
+require('virtual-global-styles');
```
chenjiahan, line 18: I'm curious why should we use `require` here..

sanyuan0704: it's not necessary to use require and i will transform it into import statement

packages/core/src/runtime/App.tsx
```diff
@@ -10,10 +10,8 @@ import { HelmetProvider } from 'react-helmet-async';
 import React, { useContext, useLayoutEffect } from 'react';
 import { Header, PageData, cleanUrl, MDX_REGEXP } from '@rspress/shared';
 import globalComponents from 'virtual-global-components';
-import 'virtual-global-styles';
-
-// eslint-disable-next-line import/no-commonjs
-const { default: Theme } = require('@theme');
+import Theme from '@theme';
+import('virtual-global-styles');
```
chenjiahan, line 14: Is dynamic import used to re-order the CSS?

sanyuan0704: The position of`import('virtual-global-styles')` should be after @theme import to avoid the latter overrides the custom global styles.

chenjiahan: ok, can we just use a sync import: `import 'virtual-global-styles';`

chenjiahan (approved):

PR #814, opened 2024-03-18
fix(theme): use react-markdown to render inline markdown

10Derozan:
## Checklist

### DIFF
--- a/.changeset/hot-timers-grab.md
+++ b/.changeset/hot-timers-grab.md
@@ -1,0 +1,5 @@
+---
+'@rspress/theme-default': patch
+---
+
+fix(theme): use react-markdown to render inline markdown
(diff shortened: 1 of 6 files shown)


sanyuan0704:
image
The size is a little large, nearly adding a react-dom package

sanyuan0704:
For text `<list>`, we can transform it as `\<list\>`

10Derozan:
> For text `<list>`, we can transform it as `\<list\>`

ok, how to test the case, regExp?

sanyuan0704:
> > For text `<list>`, we can transform it as `\<list\>`
> 
> ok, how to test the case, regExp?

Both `replaceAll` or `replace + RegExp`  can implement it.

PR #831, opened 2024-03-21
chore: remove duplicated sidebar files

sanyuan0704:
## Checklist


10Derozan (approved):

packages/theme-default/src/components/index.tsx
```diff
@@ -14,7 +14,7 @@ export { PackageManagerTabs } from './PackageManagerTabs';
 export { PrevNextPage } from './PrevNextPage';
 export { ScrollToTop } from './ScrollToTop';
 export { Search, SearchPanel } from './Search';
-export { SideBar } from './SideBar';
+export { SideBar } from './Sidebar';
```
chenjiahan, line 17: the component name is SideBar, should the file name be the same with the component name?

sanyuan0704: I think it's better to use the same name and i have updated it

chenjiahan (approved):

PR #832, opened 2024-03-21
perf: remove node-bundle-require dependency

chenjiahan:
## Summary

This PR uses the public `loadConfig` API of Rsbuild to load the config file, instead of `@modern-js/node-bundle-require`.

The `loadConfig` is based on jiti and offers better performance and less dependencies. The motivation is basically the same as https://github.com/web-infra-dev/rspress/pull/144.

After this change, Rspress will no longer depend on `@modern-js/node-bundle-require` and `esbuild`, so the install size can be reduced by 9+ mB.

## Related Issue

- https://rsbuild.dev/api/javascript-api/core#loadconfig

## Checklist

### DIFF
--- a/.changeset/angry-coins-drum.md
+++ b/.changeset/angry-coins-drum.md
@@ -1,0 +1,5 @@
+---
+"rspress": patch
+---
+
+perf: remove node-bundle-require dependency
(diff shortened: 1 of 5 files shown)


10Derozan:
nice, size reduced again.

10Derozan (approved):

chenjiahan:
> nice, size reduced again.

haha, the target is < 100mb

Issue #756, opened 2024-03-12
[Bug]: how to support monorepo's alias resolve?


wood3n:
### Version

```sh
System:
    OS: macOS 13.4
    CPU: (12) arm64 Apple M2 Pro
    Memory: 298.50 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 122.0.6261.112
```

### Details

my project dir like this. i use rspress in website directory, it depends on packages/app. it can't resolve `paths` in packages/app, so how to configure rspress to support deep nested tsconfig `paths`.

```shell
/
┣ .changeset/
┣ .husky/
┣ .vscode/
┣ node_modules/
┣ packages/
	- app
┣ website/
	- rspress.config.ts
┣ package.json
┣ pnpm-lock.yaml
┣ pnpm-workspace.yaml
```
<img width="1055" alt="image" src="https://github.com/web-infra-dev/rspress/assets/31716713/c40f9a25-d265-48c0-b6ca-684c60dca0b3">

### Reproduce link

[![Edit competent-galileo-xc2lkj](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/devbox/competent-galileo-xc2lkj?file=%2Fwebsite%2Fpackage.json&workspaceId=b94e54e5-a849-485c-ba2e-8e4a00630acd&embed=1)

### Reproduce Steps

please open codesandbox above and `pnpm run dev:web`

10Derozan:
You need to compile your app code to replace this alias.
If you want to use source code, you need to add [Source Build Plugin](https://rsbuild.dev/plugins/list/plugin-source-build) in rspress builderConfig.

wood3n:
> You need to compile your app code to replace this alias. If you want to use source code, you need to add [Source Build Plugin](https://rsbuild.dev/plugins/list/plugin-source-build) in rspress builderConfig.

I have tried Source Build, but got this error, see this reproduce: https://codesandbox.io/p/devbox/competent-galileo-xc2lkj?file=%2Fwebsite%2Fpackage.json&embed=1

<img width="986" alt="image" src="https://github.com/web-infra-dev/rspress/assets/31716713/878e3679-23e9-4439-aea9-bd28534146c3">

wood3n:
> You need add this dep to `source.include`

I tried, but still error. This error is from an inner package called `react-helmet-async`, not my code can't be compiled.

[...]

PR #833, opened 2024-03-21
fix: support source-build in mdx files

9aoy:
## Summary

fix: https://github.com/web-infra-dev/rspress/issues/756

The Rsbuild source-build plugin only applies `resolve.mainFields` and `conditionNames` in JS & TS rules, in Rspress, it also needs support in mdx files.
image

https://github.com/web-infra-dev/rsbuild/blob/main/packages/plugin-source-build/src/index.ts#L87

## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -160,8 +160,9 @@ async function createInternalBuildConfig(
     },
     tools: {
       bundlerChain(chain, { CHAIN_ID }) {
-        const swcLoaderOptions = chain.module
-          .rule(CHAIN_ID.RULE.JS)
+        const jsModuleRule = chain.module.rule(CHAIN_ID.RULE.JS);
+
+        const swcLoaderOptions = jsModuleRule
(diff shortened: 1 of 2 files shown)


10Derozan:
nice

packages/core/src/node/initRsbuild.ts
```diff
@@ -160,15 +160,21 @@ async function createInternalBuildConfig(
...
+
+        const swcLoaderOptions = jsModuleRule
           .use(CHAIN_ID.USE.SWC)
           .get('options');
 
         chain.module
           .rule('MDX')
           .type('javascript/auto')
           .test(MDX_REGEXP)
+          .resolve.merge({
```
chenjiahan, line 173: can we just inherit the `resolve` config

9aoy: can't get resolve config by `rule(js).get("resolve")`
image

chenjiahan: ok

chenjiahan (approved):

PR #834, opened 2024-03-21
fix(core): set the correct initial theme in client entry

10Derozan:
## Summary
When refreshing the page in dark mode, the logo and the icon to switch between light and dark themes will first display in light mode, and then display in dark mode after a while.

I found that the initial theme is not correct, so it will be changed. 
By the way, remove the effect to update theme in switchAppearance.

## Checklist


10Derozan:
Fix in #866

PR #840, opened 2024-03-21
fix(theme): replace angle brackets in inline markdown

10Derozan:
## Checklist

### DIFF
--- a/packages/theme-default/src/logic/utils.ts
+++ b/packages/theme-default/src/logic/utils.ts
@@ -57,6 +57,7 @@ export function renderHtmlOrText(str?: string | number | null) {
  */
 export function renderInlineMarkdown(text: string) {
   const htmlText = text
+    .replace(/</g, '&lt;')
     .replace(/\*\*(.*?)\*\*/g, '<strong>$1</strong>')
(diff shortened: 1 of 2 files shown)


packages/theme-default/src/logic/utils.ts
```diff
@@ -57,6 +57,7 @@ export function renderHtmlOrText(str?: string | number | null) {
  */
 export function renderInlineMarkdown(text: string) {
   const htmlText = text
+    .replace(/</g, '&lt;')
```
Huxpro, line 60: what about `>` and `&gt;`?

Huxpro (commented):
Have you tested on local that this works for the internal website?

packages/theme-default/src/logic/utils.ts
10Derozan: fine

10Derozan:
> Hope this works! Have you tested locally that this works for the internal website?

yeah, I have tested yet.

packages/theme-default/src/logic/utils.ts
10Derozan: only replace `<`, it works.

sanyuan0704 (approved):

PR #640, opened 2024-02-19
fix: sidebar label & toc hmr not work

sanyuan0704:
## Summary

Fix  #24  #231  

## Checklist

- [ ] I have added changeset via `pnpm run change`.
- [ ] I have updated the documentation.
- [ ] I have added tests to cover my changes.

PR #859, opened 2024-03-25
fix(core): replace flattened mdx content by rules

10Derozan:
## Summary
In #640 , we will flatten mdx content in mdx loader, and then replace-string-loader failed.
So removed it and replace in mdx loader.

## Checklist

### DIFF
--- a/packages/core/src/node/mdx/loader.ts
+++ b/packages/core/src/node/mdx/loader.ts
@@ -11,6 +11,7 @@ import {
   normalizePath,
   escapeMarkdownHeadingIds,
   flattenMdxContent,
+  applyReplaceRules,
 } from '../utils';
 import { PluginDriver } from '../PluginDriver';
 import { TEMP_DIR } from '../constants';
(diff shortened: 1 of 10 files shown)


packages/core/src/node/mdx/loader.ts
```diff
@@ -111,8 +112,10 @@ export default async function mdxLoader(
     filepath,
     alias as Record<string, string>,
   );
+  // replace content
+  const replacedContent = applyReplaceRules(flattenContent, config.replaceRules);
   // preprocessor
-  const preprocessedContent = escapeMarkdownHeadingIds(flattenContent);
+  const preprocessedContent = escapeMarkdownHeadingIds(replacedContent);
```
chenjiahan, line 118: can we add an E2E test case for this

10Derozan: yes, done

chenjiahan (approved):
LGTM 👍

PR #865, opened 2024-03-25
feat: add `localeRedirect` option to theme config

ambar:
## Summary

1. Redirecting is unexpected when sharing the specific locale.
2. It's hard to write a match function for every case, even if there is a guideline (such as https://datatracker.ietf.org/doc/html/rfc4647#section-3.4), locale codes may be too complex.

## Checklist

### DIFF
--- a/packages/document/docs/en/api/config/config-theme.mdx
+++ b/packages/document/docs/en/api/config/config-theme.mdx
@@ -574,3 +574,20 @@ export default defineConfig({
   },
 });
 ```
+
+## localeRedirect
+
+- Type: `'auto' | 'never'`
+- Default: `'auto'`
+
+Whether to redirect to the closest locale when the user visits the site, the default is `auto`, which means that the user will be redirected on the first visit. If you set it to `never`, the user will not be redirected. For example:
+
+```ts title="rspress.config.ts"
+import { defineConfig } from 'rspress/config';
+
+export default defineConfig({
+  themeConfig: {
+    localeRedirect: 'never',
+  },
+});
+```
--- a/packages/document/docs/en/api/config/config-theme.mdx
+++ b/packages/document/docs/en/api/config/config-theme.mdx
@@ -580,7 +580,7 @@ export default defineConfig({
 - Type: `'auto' | 'never'`
 - Default: `'auto'`
 
-Whether to redirect to the closest locale when the user visits the site, the default is `auto`, which means that the user will be redirected on the first visit. If you set it to `never`, the user will not be redirected. For example:
+Whether to redirect to the locale closest to `window.navigator.language` when the user visits the site, the default is `auto`, which means that the user will be redirected on the first visit. If you set it to `never`, the user will not be redirected. For example:
 
 ```ts title="rspress.config.ts"
 import { defineConfig } from 'rspress/config';
--- a/packages/document/docs/zh/api/config/config-theme.mdx
+++ b/packages/document/docs/zh/api/config/config-theme.mdx
@@ -564,7 +564,7 @@ export default defineConfig({
 - Type: `'auto' | 'never'`
 - Default: `'auto'`
 
-是否在访问网站时重定向到最接近 `window.navigator.language` 的语言，默认为 `auto` 表示首次访问时区将重定向，`never` 表示不重定向。比如:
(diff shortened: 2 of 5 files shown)


packages/shared/src/types/defaultTheme.ts
```diff
@@ -94,6 +94,11 @@ export interface Config {
    * @default false
    */
   enableScrollToTop?: boolean;
+  /**
+   * Whether to redirect to the closest locale when the user visits the site
+   * @default 'first-visit'
+   */
+  redirectToCloestLocale?: 'first-visit' | 'never';
```
chenjiahan, line 101: How about `localeRedirect?: 'auto' | 'never'`?

- auto: redirect by navigator.language
- never: do not redirect

And we can add more values in the future, such as redirect to the default language.

ambar: `first-visit` explains current behavior, may be a `always` can be added in the future.

chenjiahan: During a redirect, the redirect behaviour can only be `first-visit`, whether redirecting to the browser's language or to the default language in the configuration.

For `always`, it will cause the user to be unable to switch languages normally. 

For example, if the default language of the document is English (which means there is no `/en` path on the URL), and the current language of the browser is Chinese, if the user wants to switch to English, then the logic of `always` will making it impossible for the user to switch to English.

ambar: Makes sense, I'm open to adjustments.
I think there are at least two more strategies:
- Add a notification bar on the first visit, prompting the detection of a language that doesn't match the preference, and providing a link to confirm whether to redirect (so that users can choose not to redirect).
- Remember the current selection every time a language is chosen, and directly redirect to the user's selected language on the next visit.

chenjiahan: Agreed 😄, these two strategies will also have use cases. We can add new optional values in the future.

ambar: Rename to `localeRedirect`

packages/document/docs/zh/api/config/config-theme.mdx
```diff
@@ -559,3 +559,19 @@ export default defineConfig({
 });
 ```
 
+## localeRedirect
+
+- Type: `'auto' | 'never'`
+- Default: `'auto'`
+
+是否在访问网站时重定向到最接近的语言，默认为 `auto` 表示首次访问时区将重定向，`never` 表示不重定向。比如:
```
chenjiahan, line 567:
```suggestion
是否在访问网站时重定向到最接近 `window.navigator.language` 的语言，默认为 `auto` 表示首次访问时区将重定向，`never` 表示不重定向。比如:
```

packages/document/docs/en/api/config/config-theme.mdx
```diff
@@ -574,3 +574,20 @@ export default defineConfig({
   },
 });
 ```
+
+## localeRedirect
+
+- Type: `'auto' | 'never'`
+- Default: `'auto'`
+
+Whether to redirect to the closest locale when the user visits the site, the default is `auto`, which means that the user will be redirected on the first visit. If you set it to `never`, the user will not be redirected. For example:
```
chenjiahan, line 583:
```suggestion
Whether to redirect to the locale closest to `window.navigator.language` when the user visits the site, the default is `auto`, which means that the user will be redirected on the first visit. If you set it to `never`, the user will not be redirected. For example:
```

chenjiahan (dismissed):
LGTM 👍

packages/document/docs/zh/api/config/config-theme.mdx
```diff
@@ -559,3 +559,19 @@ export default defineConfig({
 });
 ```
 
+## localeRedirect
+
+- Type: `'auto' | 'never'`
+- Default: `'auto'`
+
+是否在访问网站时重定向到最接近 `window.navigator.language` 的语言，默认为 `auto` 表示首次访问时区将重定向，`never` 表示不重定向。比如:
```
Timeless0911, line 567: 「首次访问时区」-> 「首次访问时」

Timeless0911 (approved):

Issue #869, opened 2024-03-26
[Bug]: Section Header label doesn't read from i18n.json?


Huxpro:
### Version

```sh
System:
    OS: macOS 14.0
    CPU: (12) arm64 Apple M2 Max
    Memory: 14.98 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 123.0.6312.59
    Safari: 17.0
```

### Details

<img width="316" alt="image" src="https://github.com/web-infra-dev/rspress/assets/5563315/2f9d5ce3-504a-4897-9150-acbfae21b213">

<img width="219" alt="image" src="https://github.com/web-infra-dev/rspress/assets/5563315/a64c9dac-1927-4bd6-858c-525c84cb25ee">

Noted that I've tried restarting the dev server, as I've remembered the bug where `i18n.json` won't trigger HMR.

### Reproduce link

N/A

### Reproduce Steps

try the screenshot

ssoda01:
Would you update details about  how it is configured in `_meta.json`

Because the official example itself has a case of configuring section heading tags using `i18n.json`, perhaps this will help 🔗[Overview Page - Rspress](https://rspress.dev/guide/default-theme/overview-page.html#2-configuring-_metajson)

10Derozan:
I can not reproduce it in main branch.

Timeless0911:
fixed it in #870 https://github.com/web-infra-dev/rspress/pull/870

PR #870, opened 2024-03-27
fix(core): section header text should return i18n result

Timeless0911:
## Summary

closes: #869 

## Checklist


packages/core/src/node/runtimeModule/siteData/normalizeThemeConfig.ts
```diff
@@ -73,8 +73,17 @@ export function normalizeThemeConfig(
     const normalizeSidebarItem = (
       item: SidebarGroup | SidebarItem | SidebarDivider | SidebarSectionHeader | string,
     ): NormalizedSidebarGroup | SidebarItem | SidebarDivider | SidebarSectionHeader => {
-      // Meet the divider or section header, return directly
-      if (typeof item === 'object' && ('dividerType' in item || 'sectionHeaderText' in item)) {
+      // Meet the divider, return directly
```
chenjiahan, line 76: can we add an E2E test case for i18n.json?

Timeless0911: done

chenjiahan (approved):

chenjiahan:
👍🏻

PR #888, opened 2024-04-01
docs: add route.include hint

Timeless0911:
## Checklist


packages/document/docs/en/api/config/config-basic.mdx
```diff
@@ -307,12 +307,18 @@ import { defineConfig } from 'rspress/config';
 
 export default defineConfig({
   route: {
-    include: ['other-dir/**/*'],
+    include: ['other-dir/**/*.md'],
```
chenjiahan, line 310: usually we need to include mdx too

Timeless0911: done

sanyuan0704 (approved):

Issue #848, opened 2024-03-22
[Bug]: Excluded routes still appear in the no-config sidebars


Tyneor:
### Version

```sh
System:
    OS: Windows 10 10.0.19045
    CPU: (8) x64 Intel(R) Core(TM) i5-10310U CPU @ 1.70GHz
    Memory: 2.99 GB / 15.69 GB
  Browsers:
    Chrome: 121.0.6167.185
    Edge: Chromium (122.0.2365.92)
    Internet Explorer: 11.0.19041.3636
  npmPackages:
    rspress: ^1.0.0 => 1.15.1
```

### Details

I added the following lines to the rspress.config.ts :
```
route: {
    exclude: ["assets/**/*", "**/4-ignore.*"],
  },
```

expecting all assets in /docs and the 4-ignore.md not to have routes created for them and not appear in sidebars.

The routes are indeed not created but they still appear in the sidebar:

![image](https://github.com/web-infra-dev/rspress/assets/35744626/c9ddb029-adeb-4fe5-938f-0b47e0cd4f68)

I don't really know if this is a bug but I would say that no-config sidebars should not display non-existing pages.

### Reproduce link

https://codesandbox.io/p/devbox/minimal-route-exclude-in-sidebar-reproduction-gwzr65?file=%2Frspress.config.ts

### Reproduce Steps

1. remove `_meta.json`in docs/guide
2. add assets in docs/guide/assets/
3. add some markdown files in /docs/guide including `4-ignore.md`
4. in rspress.config.ts add 
```
route: {
    exclude: ["assets/**/*", "**/4-ignore.*"],
  },
```
5. the assets and the Ignore route still appear in the guide sidebar

10Derozan:
Yes, you are right. When fallback to readdir, we need add exclude. Welcome PR

10Derozan:
Reply to your side note:
`config` hooks can not determine the order, maybe you need to add `sidebar` to avoid internal plugin.

PR #899, opened 2024-04-02
Fix/848 Exclude config.route.exclude files from auto-generated sidebar

Tyneor:
## Summary

When no `_meta.json` is found, instead of going through all children, use `globby` to exclude files defined in `config.route.exclude`.

## Related Issue

#848 [Bug]: Excluded routes still appear in the no-config sidebars

## Checklist

- [X] I have added changeset via `pnpm run change`.
- [X] I have updated the documentation.

### DIFF
--- a/packages/plugin-auto-nav-sidebar/src/walk.ts
+++ b/packages/plugin-auto-nav-sidebar/src/walk.ts
@@ -13,7 +13,7 @@ import {
 } from '@rspress/shared';
 import { NavMeta, SideMeta } from './type';
 import { detectFilePath, extractH1Title } from './utils';
-import globby from '../../core/compiled/globby';
+import globby from 'globby';
 
 export async function scanSideMeta(
   workDir: string,
@@ -37,7 +37,6 @@ export async function scanSideMeta(
 
     const includedFiles = globby.sync('**', {
       cwd: workDir,
-      onlyFiles: false,
       ignore: [...excludedFiles],
     });
     const subItems = [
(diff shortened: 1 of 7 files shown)


packages/plugin-auto-nav-sidebar/src/walk.ts
```diff
@@ -32,7 +34,15 @@ export async function scanSideMeta(
     sideMeta = (await fs.readJSON(metaFile, 'utf8')) as SideMeta;
   } catch (e) {
     // If the `_meta.json` file doesn't exist, we will generate the sidebar config from the directory structure.
-    const subItems = await fs.readdir(workDir);
+
+    const includedFiles = globby.sync('**', {
```
sanyuan0704, line 38: Here `globby.sync` will traverse all the files and subdirectories under the directory but we only want to traverse the first level. You can change the params as follows:

```js
globby.sync('*', {
  cwd: workDir,
  ignore: excludedFiles ?? [],
  expandDirectories: false,
  onlyFiles: false,
});
```

sanyuan0704 (approved):

packages/plugin-auto-nav-sidebar/src/walk.ts
sanyuan0704: @Tyneor

Tyneor: Hello @sanyuan0704

This doesn't work for the use case `assets/**/*` I used in my original issue. With `assets/**/*` it is necessary to match the files under the directory.

I will still add `expandDirectories: false` + my second method because I think users excluding `assets/**/*` will be fine excluding `assets/**`

SoonIter:
Sorry, this PR's time was too long and had to be closed.

Now this part of the code has been completely refactored in V2, and the implementation should be simpler.

Because it is equivalent to a sidebar link, to implement it as a part of `markdown.link.checkDeadLinks` is a better choice

PR #926, opened 2024-04-08
feat(plugin-rss): sort rss items by publishing date in decreasing order

### LINKED ISSUE #911: [Feature]: RSS - sorting feed items when outputing feed

> Also I think the feed should rather be sorted in the opposite order (newest first)
> _Originally posted by @slorber in https://github.com/web-infra-dev/rspress/issues/828#issuecomment-2039673454_
            

thought both inoreader and thunderbird sort articles based on the pubDate by default when i'm adding a subscription, we should make feed items ordered in newest first when outputing feed as i'm not sure if most of the rss readers will sort items well.

1. sort feed items in descending order by pubDate
2. add sorting option for customization sorting

xc2:
## Related Issue

closes #911

## Checklist

### DIFF
--- a/.changeset/modern-olives-swim.md
+++ b/.changeset/modern-olives-swim.md
@@ -1,0 +1,5 @@
+---
+"@rspress/plugin-rss": patch
+---
+
+feat(plugin-rss): sort rss items by publishing date in decreasing order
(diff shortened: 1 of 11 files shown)


Timeless0911 (approved):

Timeless0911:
LGTM

PR #930, opened 2024-04-09
fix: add builtin rsbuild plugins before thrid party plugins

chenjiahan:
## Summary

Add builtin rsbuild plugins before third party plugins (added by `rspressConfig.builderConfig.plugins`).

- before:

Screenshot 2024-04-09 at 19 10 59

- after:

Screenshot 2024-04-09 at 19 14 23

## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -38,7 +38,10 @@ async function createInternalBuildConfig(
   isSSR: boolean,
   routeService: RouteService,
   pluginDriver: PluginDriver,
+  runtimeTempDir: string,
 ): Promise<RsbuildConfig> {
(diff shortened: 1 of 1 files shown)


packages/core/src/node/initRsbuild.ts
```diff
@@ -78,6 +81,17 @@ async function createInternalBuildConfig(
...
+      pluginReact(),
+      rsbuildPluginDocVM({
+        userDocRoot,
+        config,
+        isSSR,
+        runtimeTempDir,
+        routeService,
+        pluginDriver,
+      }),
+    ],
```
Timeless0911, line 94: It seems that the user configuration `builderPlugins` is missed.

Timeless0911: 没事了 下面有

Timeless0911 (approved):

Issue #933, opened 2024-04-10
[Bug]: Left sidebar disappeared when URL includes `.html`


chenjiahan:
### Version

```sh
"rspress": "^1.17.1"
```

### Details

- https://www.rspack.dev/blog/announcing-0.6.html

![image](https://github.com/web-infra-dev/rspress/assets/7237365/018a993d-dcd0-4344-965c-a00c272a3caa)

- https://www.rspack.dev/blog/announcing-0.6

![image](https://github.com/web-infra-dev/rspress/assets/7237365/bc74cfe1-bd59-4549-9c60-a84d03767e8d)

### Reproduce link

https://github.com/web-infra-dev/rspack/tree/main/website

### Reproduce Steps

1. cd website
2. pnpm i
3. pnpm dev

Timeless0911:
@sumy7 can you help take a look at this when you have time~

sumy7:
hi @chenjiahan @Timeless0911 
I've looked into the issue and found that the problem occurs during the process of normalizing URLs into .html links.
https://github.com/web-infra-dev/rspress/blob/main/packages/shared/src/runtime-utils/index.ts#L245
The titles containing periods are mistakenly interpreted as having file extensions. As a result, the normalized links do not correctly match the data in the sidebar.

chenjiahan:
I have upgraded to the latest version and the above bug has been fixed 👍

## Another bug

However, there is still a similar bug when using `cleanUrls: true`.

When `cleanUrls: true` is set, the generated URL will not contains `.html` suffix. But in some cases, we may still visit the page with `.html` URLs, and the sidebar will be disappeared:

![image](https://github.com/web-infra-dev/rspress/assets/7237365/87fe6392-41da-4608-9b25-c07fbce8ac70)

See:

- https://rsbuild.dev/guide/start/quick-start
- https://rsbuild.dev/guide/start/quick-start.html

It would be nice if Rspress can compatible with this case

PR #934, opened 2024-04-10
fix: url containing dots were not correctly normalized with .html extension

sumy7:
## Summary

This PR fixes an issue where URLs containing dots were not correctly appended with the `.html` extension.

```text
/guide/version-0.1
expect:
/guide/version-0.1.html
actual:
/guide/version-0.1
```

## Related Issue

https://github.com/web-infra-dev/rspress/issues/933

## Checklist

### DIFF
--- a/.changeset/spotty-pumas-travel.md
+++ b/.changeset/spotty-pumas-travel.md
@@ -1,0 +1,5 @@
+---
+"@rspress/shared": patch
+---
+
+fix: url containing dots were not correctly normalized with .html extension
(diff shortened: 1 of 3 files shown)


chenjiahan (approved):

Timeless0911:
if there exits a url like `/foo/bar.png`, it will be added a `.html` in current logic? I'm not sure if there is such a scenario, may be we need to know what is the original extension judgment logic for?

sumy7:
> if there exists a url like `/foo/bar.png`, it will be added a `.html` in current logic? I'm not sure if there is such a scenario, may be we need to know what is the original extension judgment logic for?
> 
> 
> 
> @sanyuan0704 any context about this logic?

you are right. We should check all the usages of this function. Or add a flag to control this action.

sumy7:
I have reviewed all the places where the function `normalizeHref` is used. It appears to be used only for organizing href parameters or handling article links, and there is no case where it handles such as `.png` suffixes. I will reopen this PR for now, but further confirmation is needed from @sanyuan0704 .

sanyuan0704 (approved):

PR #940, opened 2024-04-11
fix: avoid unexpected changes to searchHooks

Timeless0911:
## Summary

Previously in `siteDataVMPlugin` we modify the siteData object to deal with some issues?

But it's a shallow copy, the config will be used in other runtime module like `searchHookVMPlugin`.

In below codes, searchHooks has been set undefined, which cause `searchHookVMPlugin` inject an empty module.

https://github.com/web-infra-dev/rspress/blob/8b0649b731f5751947e70be0c8245d1f92dd5bf6/packages/core/src/node/runtimeModule/siteData/index.ts#L31

https://github.com/web-infra-dev/rspress/blob/8b0649b731f5751947e70be0c8245d1f92dd5bf6/packages/core/src/node/runtimeModule/siteData/index.ts#L105

https://github.com/web-infra-dev/rspress/blob/8b0649b731f5751947e70be0c8245d1f92dd5bf6/packages/core/src/node/runtimeModule/siteData/index.ts#L120-L122

## Checklist

### DIFF
--- a/packages/core/src/node/runtimeModule/siteData/index.ts
+++ b/packages/core/src/node/runtimeModule/siteData/index.ts
@@ -29,6 +29,14 @@ function deletePriviteKey<T>(obj: T): T {
 export async function siteDataVMPlugin(context: FactoryContext) {
   const { config, alias, userDocRoot, routeService, pluginDriver } = context;
   const userConfig = config;
+  // prevent modify the origin config object
+  const tempSearchObj = Object.assign({}, userConfig.search);
+
+  // searchHooks is a absolute path which may leak information
+  if (tempSearchObj) {
+    tempSearchObj.searchHooks = undefined;
+  }
+
   const replaceRules = userConfig?.replaceRules || [];
   // If the dev server restart when config file, we will reuse the siteData instead of extracting the siteData from source files again.
   const domain =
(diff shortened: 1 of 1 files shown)


10Derozan (dismissed):

packages/core/src/node/runtimeModule/siteData/index.ts
```diff
@@ -28,7 +28,8 @@ function deletePriviteKey<T>(obj: T): T {
 
 export async function siteDataVMPlugin(context: FactoryContext) {
   const { config, alias, userDocRoot, routeService, pluginDriver } = context;
-  const userConfig = config;
+  // deep clone to avoid unexpected changes to objects
+  const userConfig = JSON.parse(JSON.stringify(config));
```
chenjiahan, line 32: Will the config contain some properties that can not be serialized?

chenjiahan: such as RegExp or function

Timeless0911: see the latest code, just use Object.assign to prevent the search object being modified

chenjiahan: LGTM

chenjiahan (approved):

PR #955, opened 2024-04-14
fix(core): normalizeRoutePath remove only known extensions

### LINKED ISSUE #954: [Bug]: links with `.` (periods) are reported as broken
### Version

```sh
System:
    OS: macOS 14.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 39.19 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 123.0.6312.123
    Safari: 17.1
  npmPackages:
    rspress: ^1.18.1 => 1.18.1
```

### Details

When using `typedoc`, it can generate names of files with periods in them like this:
- `classes/plugins.OutputPlugin`
- `classes/plugins.LoggerPlugin`
then the filename gets put into the route path and despite the links working, they will get reported as being broken when using `markdown.checkDeadLinks` option in the config

### Reproduce link

https://github.com/callstack/repack/tree/docs/migration

### Reproduce Steps

1. `pnpm install`
2. `cd new-website`
3. `pnpm start`

jbroma:
## Summary

- remove only known extensions from route paths
- removed redundant `normalizeRoutePath` from `routeData`
- expose `normalizeRoutePath` from `RouteService`

## Related Issue

fix #954 

## Checklist


Timeless0911:
LGTM

Timeless0911 (dismissed):

packages/core/src/node/mdx/remarkPlugins/checkDeadLink.ts
```diff
@@ -66,9 +65,9 @@ export const remarkCheckDeadLinks: Plugin<
       }
 
       if (!url.startsWith('http') && !url.startsWith('https')) {
-        const normalizeUrl = normalizeRoutePath(
+        const { routePath: normalizeUrl } = routeService.normalizeRoutePath(
           // fix: windows path
-          url?.split(path.sep).join('/')?.split('#')[0],
+          url.split(path.sep).join('/')?.split('#')[0],
```
jbroma, line 70: FYI, I've removed optional operator here as from TS definitions it didn't seem to be needed

sanyuan0704:
Why do you want to put the implement of normalizeRoutePath into `RouteService` instead of acting as a universal util function? @jbroma

jbroma:
> Why do you want to put the implement of normalizeRoutePath into `RouteService` instead of acting as a universal util function? @jbroma

@sanyuan0704 The reason `normalizeRoutePath` was put inside of `RouteService` was to access the list of extensions from the user supplied settings. We could keep it outside but then we would need to rely on default extensions only.

sanyuan0704:
> > Why do you want to put the implement of normalizeRoutePath into `RouteService` instead of acting as a universal util function? @jbroma
> 
> @sanyuan0704 The reason `normalizeRoutePath` was put inside of `RouteService` was to access the list of extensions from the user supplied settings. We could keep it outside but then we would need to rely on default extensions only.

@jbroma You can add extensions param in `normalizeRoutePath` and call it in the private method of RouteService.

jbroma:
@sanyuan0704 oh right, that makes sense, I'll rework this PR so it's not a breaking change like this, forgot this are not internal APIs, thanks for spotting that!

jbroma:
@sanyuan0704 upon closer inspection these functions were not exposed as public API outside of core package so they wouldn't be breaking (only `normalizeRoutePath` from `rspress/runtime` is public)

I've brough back `normalizeRoutePath` outside of the `RouteService` but also made the previously private `#normalizeRoutePath` public so that we can use the extensions from `RouteService` inside of `checkDeadLinks`.

Let me know whether this is ok now

sanyuan0704 (approved):

Issue #947, opened 2024-04-12
[Feature]: Separate search index by version


jbroma:
### What problem does this feature solve?

When searching through documentation, I would like to only display results relevant to the currently selected version.
Right now, all versions get combined into one search index.

Splitting the search index into separate entities by version would also reduce size needed to download upon opening the search panel.

### What does the proposed API look like?

To make it non-breaking, it could be introduced as an optional prop to `config.search` like this:

```ts
//rspress.config.ts
export default defineConfig({
  search: {
    versioned: true,
  },
});
```

sanyuan0704:
> @sanyuan0704 I see the PR got merged, but the docs were not updated to reflect this change, I think it's important to mention that somewhere.

@jbroma Would you like to add some docs to reflect this change?

jbroma:
sure thing, I'll add the english part 👍

PR #962, opened 2024-04-15
docs: versioned search

jbroma:
## Summary

added english documentation for `versioned` search.

## Related Issue

#947 

## Checklist

### DIFF
--- a/packages/document/docs/en/api/config/config-basic.mdx
+++ b/packages/document/docs/en/api/config/config-basic.mdx
@@ -233,6 +233,19 @@ export async function onSearch(query: string) {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Timeless0911:
Oh, the document of `search` seems to be missing a lot of parameters, we will have to find time to complete it then, and update the Chinese documentation here by the way.

jbroma:
@Timeless0911 yeah I agree 💯, it seems like `RemoteProvider` is something not exposed at all

Timeless0911:
@sanyuan0704 The config about `search` in document is only contain a `searchHooks` options, should we add all config which is implemented in code?

Timeless0911 (approved):

PR #963, opened 2024-04-16
feat(api-docgen): support more config to init parser instance

Timeless0911:
## Summary

If the parser is `react-docgen-typescript`, the `withDefaultConfig` method is used by default to create a parser instance. If `tsconfigPath` or `compilerOptions` is configured, the `withCompilerOptions` and `withCustomConfig` methods are used to create the parser instance respectively. See [Custom Parsers](https://github.com/styleguidist/react-docgen-typescript/blob/b7ea0a235efb7c78a1158ca12d864de2bc2ee30e/README.md#usage) for details.

## Checklist

### DIFF
--- a/packages/plugin-api-docgen/package.json
+++ b/packages/plugin-api-docgen/package.json
@@ -45,7 +45,8 @@
   "peerDependencies": {
     "react": ">=17",
     "react-router-dom": "^6.8.1",
-    "@rspress/core": "^1.0.2"
+    "@rspress/core": "^1.0.2",
+    "typescript": "^5"
   },
   "files": [
     "dist",
--- a/packages/plugin-api-docgen/package.json
+++ b/packages/plugin-api-docgen/package.json
@@ -48,6 +48,11 @@
     "@rspress/core": "^1.0.2",
     "typescript": "^5"
   },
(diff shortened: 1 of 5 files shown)


Timeless0911:
!canary

packages/plugin-api-docgen/package.json
```diff
@@ -45,7 +45,8 @@
   "peerDependencies": {
     "react": ">=17",
     "react-router-dom": "^6.8.1",
-    "@rspress/core": "^1.0.2"
+    "@rspress/core": "^1.0.2",
+    "typescript": "^5"
```
chenjiahan, line 49: should this be optional?

packages/plugin-api-docgen/src/types.ts
```diff
@@ -1,5 +1,6 @@
 import type { PageIndexInfo } from '@rspress/shared';
 import type { ParserOptions } from 'react-docgen-typescript';
+import * as ts from 'typescript';
```
chenjiahan, line 3: can we just import type

chenjiahan (approved):

Timeless0911:
!canary

PR #968, opened 2024-04-17
fix: checkDeadLinks when mdxRs is set to false

### LINKED ISSUE #965: [Bug]: CheckDeadLinks broken when `markdown.mdxRs` set to false
### 版本信息

```sh
-
```

### 问题详情

In this PR #955 , we modify the checkLinks function for remove only known extension, but in below code (checkDeadLinks logic in `mdxRs: false`)

https://github.com/web-infra-dev/rspress/blob/221da5e8ead5ec59a5f4fdc61b2a1dc534483377/packages/core/src/node/mdx/remarkPlugins/checkDeadLink.ts#L67-L76

The `internalLinks` variable will contain `.html` suffix， which caused the error:

image

In former logic, we remove all `.*` extensions, but now we only remove `DEFAULT_PAGE_EXTENSIONS = ['js', 'jsx', 'ts', 'tsx', 'md', 'mdx']`. 

Maybe we need to add a logic to remove the `.html` suffix after line 72?

### 复现链接

https://github.com/web-infra-dev/rspress/blob/main/packages/document/package.json#L15

### 复现步骤

1. `pnpm i`
2. modify https://github.com/web-infra-dev/rspress/blob/221da5e8ead5ec59a5f4fdc61b2a1dc534483377/packages/document/rspress.config.ts#L17

add `mdxRs: false` config

```ts
markdown: {
    checkDeadLinks: true,
    mdxRs: false,
},
```
3. `pnpm dev`

jbroma:
## Summary

handle `.html` extension separately, because `extensions` can be overridden via config, but it's always present in the output.

## Related Issue

closes #965 

## Checklist

### DIFF
--- a/packages/core/src/node/route/RouteService.ts
+++ b/packages/core/src/node/route/RouteService.ts
@@ -69,6 +69,7 @@ export const normalizeRoutePath = (
   )
     // remove the extension
     .replace(new RegExp(`\\.(${extensions.join('|')})$`), '')
+    .replace(/\.html$/, '')
     .replace(/\/index$/, '/');
 
   // restore the trail slash


Timeless0911:
LGTM

Timeless0911 (approved):

Issue #971, opened 2024-04-18
[Bug]: Page crashes when using <Tabs /> with a single <Tab /> as children


chenjiahan:
### Version

```sh
Rspress v1.18.1
```

### Details

Page crashes when using <Tabs /> with a single <Tab /> as children, for example:

```jsx
<Tabs>
  <Tab label="foo">foo</Tab>
</Tabs>
```

Related code: https://github.com/web-infra-dev/rspress/blob/0625b783ca81ea04436320c5a4c6eb837a07944c/packages/theme-default/src/components/Tabs/index.tsx#L57-L60

The `children` is JSX when there is only a single item. Maybe we should use `React.Children.map()` instead of `children.map`

> this issue war originally reported by @colinaaa ❤️

### Reproduce link

https://github.com/web-infra-dev/rspress

### Reproduce Steps

1. add the code
2. pnpm dev

chenjiahan:
I will take this

PR #972, opened 2024-04-18
fix: page crashes if Tabs has single Tab child

chenjiahan:
## Summary

Fix page crashes if Tabs has single Tab child.

## Related Issue

https://github.com/web-infra-dev/rspress/issues/971

## Checklist


Timeless0911:
If we write like this

```jsx
<Tabs tabContainerClassName="tabs-b">
  <Tab label="label2">content2</Tab>
  <Tab label="label3">content3</Tab>
</Tabs>
```

It contains a newline character in the middle of two `<Tab>`

image

We should filter the `children` array to remove the '\n' element

Timeless0911 (approved):

PR #973, opened 2024-04-18
fix: Tabs component children contains linebreak or with no tab

Timeless0911:
## Summary

fix tabs component children contains linebreak or with no tab

## Checklist

### DIFF
--- a/e2e/tests/tabs-component.test.ts
+++ b/e2e/tests/tabs-component.test.ts
@@ -38,5 +38,18 @@ test.describe('tabs-component test', async () => {
     const contentBText = await page.evaluate(node => node?.innerHTML, contentB);
     expect(tabBText).toContain('label2');
     expect(contentBText).toContain('content2');
+    const notSelected = await page.$$eval(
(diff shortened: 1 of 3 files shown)


e2e/tests/tabs-component.test.ts
```diff
@@ -38,5 +38,15 @@ test.describe('tabs-component test', async () => {
     const contentBText = await page.evaluate(node => node?.innerHTML, contentB);
     expect(tabBText).toContain('label2');
     expect(contentBText).toContain('content2');
+    const notSelected = await page.$$('.tabs-b div.not-selected_fa768');
```
chenjiahan, line 41: `.not-selected_fa768` is an unstable selector and may be broken in the future iterations.

Can we use another selector

Timeless0911: ok, theoretically hash should be stable.

I changed the way of selecting, pls check it again

chenjiahan (approved):

PR #977, opened 2024-04-18
feat(playgroud): pass source path to custom render

Timeless0911:
## Checklist


Timeless0911:
!canary

packages/plugin-playground/static/global-components/Playground.tsx
```diff
@@ -13,6 +13,7 @@ type Direction = 'horizontal' | 'vertical';
 export interface PlaygroundProps extends HTMLAttributes<HTMLDivElement> {
   code: string;
   language: string;
+  source?: string;
```
10Derozan, line 16: comment?

PR #979, opened 2024-04-18
fix(modern-plugin): improve ignore for auto sidebar

10Derozan:
## Summary
Fix 
image
If we add some assets in public.
## Checklist

### DIFF
--- a/packages/modern-plugin-rspress/src/lanchDoc.ts
+++ b/packages/modern-plugin-rspress/src/lanchDoc.ts
@@ -52,10 +52,11 @@ export async function launchDoc({
     prefix: string,
   ): Promise<SidebarGroup> => {
     const base = join(root, lang);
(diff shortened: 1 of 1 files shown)


packages/modern-plugin-rspress/src/lanchDoc.ts
```diff
@@ -85,7 +87,7 @@ export async function launchDoc({
           const directoryCwd = join(cwd, directory);
           const hasIndex =
             (
-              await fastGlob(['index.*'], {
+              await fastGlob(source, {
```
Timeless0911, line 90: 这里差别咋这么大

10Derozan: Reset yet

Timeless0911 (approved):

PR #980, opened 2024-04-18
✨ feat(route): Update home page title with frontmatter titleSuffix

h7ml:
Updated the logic for generating the home page title to include the titleSuffix from frontmatter if available, using the format ${mainTitle} - ${frontmatter.titleSuffix}, or defaulting to mainTitle if titleSuffix is not present.

## Checklist

### DIFF
--- a/packages/document/docs/en/index.md
+++ b/packages/document/docs/en/index.md
@@ -1,5 +1,6 @@
 ---
 pageType: home
+titleSuffix: - Rspack-based Static Site Generator
 
 hero:
   name: Rspress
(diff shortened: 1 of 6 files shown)


chenjiahan:
Sry but these two files do not seem to have been updated.

https://github.com/web-infra-dev/rspress/blob/main/packages/document/docs/zh/index.md
https://github.com/web-infra-dev/rspress/blob/main/packages/document/docs/en/index.md

packages/document/docs/en/index.md
```diff
@@ -1,5 +1,6 @@
 ---
 pageType: home
+titleSuffix: - Rspack-based Static Site Generator
```
chenjiahan, line 3: The `-` breaks yaml and we need to quote it..

```suggestion
titleSuffix: '- Rspack-based Static Site Generator'
```

PR #987, opened 2024-04-19
fix: crash when page not found & using multiVersion

### LINKED ISSUE #986: [Bug]: crash when page not found and using multiVersion
### Version

```sh
System:
    OS: macOS 14.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 39.19 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 123.0.6312.123
    Safari: 17.1
  npmPackages:
    rspress: 1.18.2
```

### Details

when using multiVersion, page crashes when accessing non-existent route instead of displaying 404. 

### Reproduce link

n/a

### Reproduce Steps

1. setup multiVersion
2. access non-existent route
3. observe page crash instead of 404

jbroma:
## Summary

When using `multiVersion`, the `useNav` hook defaults to 'default' version when version from `useVersion` is empty. We should first check for defaultVersion in `siteData.multiVersion.default` before going with `default` as a key for nav.

Edit: as this happens on 404 page only, it's better to supply `defaultVersion` from `siteData.multiVersion.default` in `initPageData`, similar as it is done for lang. This ensured that:
- [x] - page doesn't crash in `multiVersion` setup
- [x] - when using `search.versioned`, fetching search index doesn't fail, and fetches default version index
- [x] - default version is displayed on 404 page version dropdown

## Related Issue

closes #986 

## Checklist


Timeless0911 (approved):

Timeless0911:
image

Seems github API error, but only in this PR, i will try to close this PR first and open another one

PR #998, opened 2024-04-21
chore(deps): unify fs-extra version

SoonIter:
## Summary

unify fs-extra version

image

## Related Issue
none

## Checklist


10Derozan:
which problem may be caused before unify?

SoonIter:
> which problem may be caused before unify?
Not at present, but it can reduce the size of the package.

chenjiahan (approved):

chenjiahan (approved):

chenjiahan:
Nice catch 👍🏻

Issue #574, opened 2024-01-24
[Feature]:  integration with twoslash which display ts info in code blocks


sanyuan0704:
### What problem does this feature solve?

See the details in  https://github.com/shikijs/twoslash.

### What does the proposed API look like?

Now we support shiki highlighter in rspress-plugin-shiki. We can

SoonIter:
In Rspress V2, we default to using shiki for syntax highlighting, so this issue can be resolved after enable twoslash plugin.

I have not tested its behavior yet. So pr welcome

Karibash:
`@shikijs/twoslash` provides several renderers with different approaches to displaying type information.

- rendererClassic: a simple renderer that only displays type information.
- rendererRich: extends the output to include both type information and TSDoc comments.
- rendererFloatingVue: uses [FloatingVue](https://floating-vue.starpad.dev/)
 to show popups. It is implemented as a wrapper around `rendererRich` and is also the renderer used by `@shikijs/vitepress-twoslash`.

Compared to `rendererFloatingVue`, both `rendererClassic` and `rendererRich` provide a less sophisticated UI/UX since they rely only on CSS for popups. While these renderers do ship with CSS, the styles cannot be used directly in Rspress. To achieve a UI/UX similar to VitePress, you would likely need to create a plugin such as `rspress-twoslash`.

It is also worth noting that Twoslash builds HTML directly from its AST, which may make it difficult to integrate popup rendering smoothly with React. The `rendererFloatingVue` manages this by leveraging `<template>` tags for integration, but it is unclear whether a similar approach would work seamlessly in React.

Because of this, contributing in this area is quite challenging for external contributors. It requires knowledge of Twoslash and HTML AST, expertise in styling for proper integration with Rspress, and also the ability to evaluate and adopt a React library for popup rendering.

I attempted to address this by using [FloatingUI](https://floating-ui.com/docs/getting-started) to display popups.However, [my solution](https://github.com/praha-inc/byethrow/blob/120d2822459ac7497986deadb8247a937fa923b3/website/plugins/twoslash/popup.tsx) forcefully manipulates the DOM tree managed by React, which is admittedly not a good implementation. Ideally, there would be a way to integrate with React as smoothly as FloatingVue does with `<template>`, but unfortunately, I didn’t know how to achieve that.

Karibash:
I made some progress on this.
I refactored the code and managed to display popups using Web Components:
[...]

PR #1001, opened 2024-04-21
chore(deps): update markdown-related dependencies

colinaaa:
## Summary

I'm trying out #574. The `@shikijs/rehype` uses a higher version of
`unified` than Rspress. So I decided to first upgrade the dependencies of
Rspress.

Upgrade `@mdxjs/*`, `hast-*`, `mdast-util-*`, `rehype-*`, `remark-*`,
`unist-util-*` and `unified` to latest.

Please note that these are major bumps with breaking changes(mostly
typings). The `remarkPlugins/toc.ts` and `remarkPlugins/loader.ts` have
been updated for the correct types.

I've tried running Rspack and Rsbuild documentation with the upgraded
Rspress. Everything looks just fine!

## Related Issue

#574

## Checklist


packages/core/package.json
```diff
@@ -40,32 +40,32 @@
   },
   "dependencies": {
     "@loadable/component": "5.16.4",
-    "@mdx-js/loader": "2.3.0",
-    "@mdx-js/mdx": "2.3.0",
-    "@mdx-js/react": "2.3.0",
+    "@mdx-js/loader": "3.0.1",
+    "@mdx-js/mdx": "3.0.1",
```
chenjiahan, line 44: I'm worried that mdx v3 and remark v15 contains some breaking changes and may break the user code / plugins.

chenjiahan: I believe the [mdx-rs](https://github.com/web-infra-dev/mdx-rs) of Rspress use the mdx v2 syntax, @sanyuan0704 can you confirm this?

colinaaa: Looks like [mdx v3 updates the rehype and remark](https://mdxjs.com/migrating/v3/#update-plugins), which contains breaking changes mainly in typings. This is indeed a breaking change for Rspress users, but migration is easy. Just update the rehype and remark plugins.

Issue #1004, opened 2024-04-22
[Bug]: Can't resolve 'stream' in react-dom


topazur:
### Version

```sh
mac 14
chrome 123.0.6312.124
pnpm 7
```

### Details

error   Compile error: 
Failed to compile, check the errors for troubleshooting.
./node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js
  × Resolve error: Can't resolve 'stream' in '/node_modules/.pnpm/react-dom@18.2.0_react@18.2.0/node_modules/react-dom/cjs'
    ╭─[11:1]
 11 │     'use strict';
 12 │     var React = require('react');
 13 │     var stream = require('stream');
    ·                  ─────────────────
 14 │     var ReactVersion = '18.2.0';
 15 │     var ReactSharedInternals = React.__SECRET_INTERNALS_DO_NOT_USE_OR_YOU_WILL_BE_FIRED;
    ╰────

Tip: "stream" is a built-in Node.js module and cannot be imported in client-side code.
Check if you need to import Node.js module. If needed, you can use "@rsbuild/plugin-node-polyfill".

### Reproduce link

无

### Reproduce Steps

- pnpm add ag-grid-react

- a.jsx
  ```jsx
  import { AgGridReact } from 'ag-grid-react'
  export default () => {
    return (<AgGridReact />)
  }
  ```

- in mdx file => `import Demo from './a.jsx'; OR <code src='./a.jsx' />`

PR #1019, opened 2024-04-25
fix: client compile error when set react-dom/server alias

Timeless0911:
## Related Issue

closes: #1004 

## Checklist

### DIFF
--- a/packages/core/src/node/utils/detectReactVersion.ts
+++ b/packages/core/src/node/utils/detectReactVersion.ts
@@ -23,7 +23,7 @@ export async function detectReactVersion(): Promise<number> {
   return DEFAULT_REACT_VERSION;
 }
 
-export async function resolveReactAlias(reactVersion: number) {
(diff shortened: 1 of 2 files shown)


packages/core/src/node/utils/detectReactVersion.ts
```diff
@@ -65,5 +65,14 @@ export async function resolveReactAlias(reactVersion: number) {
       }
     }),
   );
+
+  // react-dom/server should resolve to browser when targets is web
+  if (alias?.['react-dom/server'] && !isSSR) {
```
chenjiahan, line 70: Can we change the resolve options of `ResolverFactory.createResolver` to resolve the browser bundles correctly?

Timeless0911: done，this logic only handles react-related dependencies, so modifying `conditioalNames` should have no impact.

chenjiahan (approved):

PR #1020, opened 2024-04-25
fix: support use same name file as sidebar index when no _meta.json

Timeless0911:
## Summary

1. If there exists a file with the same name of the directory folder, we use it for sidebar index
2. if 1, the sidebar label will be the H1 title of the same name file 

## Checklist


e2e/fixtures/auto-nav-sidebar-no-meta/doc/api/rspress-commands.mdx
```diff
@@ -0,0 +1,5 @@
+# 命令
```
chenjiahan, line 1: can we use English by default in e2e cases

chenjiahan (dismissed):

e2e/fixtures/auto-nav-sidebar-no-meta/doc/api/rspress-commands.mdx
Timeless0911: done

chenjiahan (approved):

Issue #995, opened 2024-04-20
[Bug]: Changing language on 404 page does nothing


jbroma:
### Details

When on 404 page, changing language does nothing and only appends new lang to the URL. 
Similar thing happens when using multiVersion and changing versions while on 404 page.

The question here is: what should be the desired behaviour?

### Reproduce link

https://rspress.dev/404 (or any non-existent route)

### Reproduce Steps

1. Go to the 404 page on `rspress.dev`
2. Change language
3. Observe 'zh' being added to the URL indefinetly, after few times it looks like this:

<img width="509" alt="image" src="https://github.com/web-infra-dev/rspress/assets/20626315/e31fe888-eb39-4104-a7b3-a40ef6a08a69">

Timeless0911:
The desired behavior is that the 404 page does not have multi-version and language switching, and there should be some bugs now when we switching.

PR #1023, opened 2024-04-26
fix(theme-default): changing language and version on 404 page does nothing

shulaoda:
## Related Issue

closes: #995 

## Checklist

### DIFF
--- a/packages/shared/src/runtime-utils/index.ts
+++ b/packages/shared/src/runtime-utils/index.ts
@@ -124,10 +124,11 @@ export function replaceLang(
   },
   base = '',
   cleanUrls = false,
+  isPageNotFound = false,
 ) {
   let url = removeBase(rawUrl, base);
   // rspress.dev/builder + switch to en -> rspress.dev/builder/en/index.html
-  if (!url) {
+  if (!url || isPageNotFound) {
     url = cleanUrls ? '/index' : '/index.html';
   }
 
@@ -179,10 +180,11 @@ export function replaceVersion(
   },
   base = '',
   cleanUrls = false,
+  isPageNotFound = false,
 ) {
   let url = removeBase(rawUrl, base);
   // rspress.dev/builder + switch to en -> rspress.dev/builder/en/index.html
-  if (!url) {
+  if (!url || isPageNotFound) {
     url = cleanUrls ? '/index' : '/index.html';
   }
   let versionPart = '';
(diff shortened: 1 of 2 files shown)


Timeless0911:
Can you help modify this logic to let 404 page back to homepage of current lang?

https://github.com/web-infra-dev/rspress/blob/588c9f22a288922a96df4a2e9551f627e1dcf287/packages/theme-default/src/layout/NotFountLayout/index.tsx#L32

shulaoda:
> Can you help modify this logic to let 404 page back to homepage of current lang?

Yes, it's already done.

packages/theme-default/src/layout/NotFountLayout/index.tsx
```diff
@@ -1,21 +1,43 @@
...
     location.pathname.includes(`/${defaultLang}/`)
   ) {
     const redirectUrl = location.pathname.replace(`/${defaultLang}/`, '/');
     window.location.replace(redirectUrl);
     return <></>;
   }
 
+  const root = useMemo(() => {
+    if (isBrowser) {
+      const currentLang = location.pathname
```
Timeless0911, line 23: when we enable multiversion, the url will be like `/v2/zh/index.html`, this logic have little problem?

May be we can use `usePageData()` hook, to get the `page.lang` for currentLang? I'm not sure if this will work. You can check to see if there is any relevant implementation of getting the current language in the entire repo.

shulaoda: Both of `page.lang` and `useLang` can't work. Maybe I should use the logic above `includes` or make more judgments on version in the changed code. What's your advice?

Timeless0911: https://github.com/web-infra-dev/rspress/blob/faf6bcf8aae5394db395773da27fb433b33d0b0a/packages/theme-default/src/layout/Layout/index.tsx#L119

pass the currentLang in props here?

shulaoda: Should we consider the current version? Go back to the home of the current version in the current language or just the default version in the current language?

Timeless0911: the current version in the current language is better

shulaoda: > pass the currentLang in props here?

This doesn't work. I already have some ideas, and I will open a new PR to solve it.

Timeless0911 (approved):

Issue #891, opened 2024-04-01
[Feature]: Support tooltip when hover Aside item


10Derozan:
### What problem does this feature solve?

When hover in item, we can show a tooltip for overflow.
![image](https://github.com/web-infra-dev/rspress/assets/50694858/8042469b-39ef-49ce-b121-9ebf073d69ac)

### What does the proposed API look like?

No API.

ArnavK-09:
# 🙋‍♂️

PR #1025, opened 2024-04-26
feat: support tooltip when hover aside item

shulaoda:
## Related Issue

closes: #891 

## Checklist

### DIFF
--- a/packages/theme-default/src/logic/utils.ts
+++ b/packages/theme-default/src/logic/utils.ts
@@ -37,17 +37,29 @@ export function renderHtmlOrText(str?: string | number | null) {
   return isHtml(str) ? htmr(str) : str;
 }
 
+const CODE_TEXT_PATTERN = /`(.*?)`/g;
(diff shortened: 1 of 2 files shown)


packages/theme-default/src/logic/utils.ts
```diff
@@ -51,3 +51,10 @@ export function renderInlineMarkdown(text: string) {
     .replace(/`(.*?)`/g, '<code>$1</code>');
   return renderHtmlOrText(htmlText);
 }
+
+export function parseInlineMarkdownText(mdx: string) {
+  return mdx
```
Timeless0911, line 56: We can use const to store these regular expressions, so that they are easier to maintain later, including the above `renderInlineMarkdown` functions.

By the way, I noticed in former PR, `renderInlineMarkdown` functions has been modified but the param comment not updated, you can updated it in this PR

Timeless0911:
Much appreciate for your awesome contribution!

Timeless0911 (approved):

Issue #1026, opened 2024-04-26
[Bug]: `markdown.highlightLanguages` config does not take effect


Timeless0911:
### Version

```sh
-
```

### Details

In Rspress doc https://rspress.dev/api/config/config-build#markdownhighlightlanguages

>Each language config supports both string and [string, string] formats. In the array format, the former is the alias of the language, and the latter is the full name of the language. You can go to the [file list](https://github.com/react-syntax-highlighter/react-syntax-highlighter/tree/master/src/languages/prism) to view all supported language full names.

Here can be understood as we can configure aliases to map some unsupported languages to the languages supported by prism( etc `['mdx', 'tsx']`), or we can also make an abbreviation for a certain language ( etc `['oc', 'objectivec']`)

But now there should be some problems in existing implementation, below code block are not highlighted.

```oc
#import <Foundation/Foundation.h>

int main(int argc, const char * argv[]) {
    @autoreleasepool {
        NSLog(@"Hello, Objective-C!");
    }
    return 0;
}
```

```mdx
<Alert type="info">This is a info alert</Alert>
```

### Reproduce link

https://github.com/web-infra-dev/rspress/tree/main/e2e/fixtures/basic

### Reproduce Steps

1. cd https://github.com/web-infra-dev/rspress/tree/main/e2e/fixtures/basic
2. modify `rspress.config.ts`
```ts
import * as path from 'path';
import { defineConfig } from 'rspress/config';

export default defineConfig({
  root: path.join(__dirname, 'doc'),
  markdown: {
    highlightLanguages: [
      ['oc', 'objectivec'],
      ['mdx', 'tsx'],
    ],
  },
});

```

`['mdx', 'tsx']` is the `DEFAULT_HIGHLIGHT_LANGUAGES` in https://github.com/web-infra-dev/rspress/blob/e42143357e9071bbc2e1e1821d29d4e66f1efd45/packages/shared/src/runtime-utils/index.ts#L11-L26

You can also try `['mdx', 'md']`, may be also check whether there are any problems with the logic of merging and deduplication highlight languages.

3. add some code blocks like below in `index.mdx`

![image](https://github.com/web-infra-dev/rspress/assets/50201324/8f56d387-115a-40b6-ae46-4b88da7dfcbf)

(since github will transform the raw code, so screenshot here)

4. run `pnpm dev`

6. see if the code is highlighted

PR #1041, opened 2024-04-29
fix: markdown highlight languages

shulaoda:
## Related Issue
#1026 

## Checklist

### DIFF
--- a/packages/core/src/node/runtimeModule/prismLanguages.ts
+++ b/packages/core/src/node/runtimeModule/prismLanguages.ts
@@ -5,14 +5,20 @@ import { FactoryContext, RuntimeModuleID } from '.';
 export async function prismLanguageVMPlugin(context: FactoryContext) {
   const { config } = context;
   const { highlightLanguages = [] } = config.markdown || {};
+  const aliases: Record<string, string[]> = {};
   const languageMeta = uniqBy(
     [...DEFAULT_HIGHLIGHT_LANGUAGES, ...highlightLanguages].map(language => {
-      const [alias, name] = Array.isArray(language)
-        ? language
-        : [language, language];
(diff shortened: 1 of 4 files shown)


packages/core/src/node/runtimeModule/prismLanguages.ts
```diff
@@ -5,14 +5,20 @@ import { FactoryContext, RuntimeModuleID } from '.';
...
     [...DEFAULT_HIGHLIGHT_LANGUAGES, ...highlightLanguages].map(language => {
-      const [alias, name] = Array.isArray(language)
-        ? language
-        : [language, language];
+      const isArray = Array.isArray(language);
+      const [alias, name] = isArray ? language : [language, language];
+
+      if (isArray) {
+        const temp = aliases[name] || (aliases[name] = []);
+        temp.push(alias);
```
Timeless0911, line 16: Users may repeatedly configure existing default values, we may need to determine whether the element already exists in the array, but it will have no impact on subsequent logic.

eg

``` ts
markdown: {
    highlightLanguages: [
      ['js', 'javascript'],
      ['ts', 'typescript'],
    ],
  },
```

result in 

```ts
aliases:  {
  javascript: [ 'js', 'js' ],
  typescript: [ 'ts', 'ts' ],
  tsx: [ 'jsx', 'mdx' ],
  'xml-doc': [ 'xml' ],
  markdown: [ 'md' ]
}
```

Timeless0911:
Awesome! Can you add an e2e fixtures named `prism-sytax-highlighter`, and write a simple test for this?

Timeless0911 (approved):

Timeless0911:
Thank you~

PR #1072, opened 2024-05-09
fix: globalStyles being shaked when config some sideEffects

Timeless0911:
## Summary

In a module project, the sideEffects usually be

```
"sideEffects": [
  "**/*.css",
  "**/*.less",
  "**/*.sass",
  "**/*.scss"
],
```
 when we use module doc plugin, it will build a web project, the sideEffects config will influence the tree-shaking behaviour of global styles virtual module, which caused some styles broken in the current version of rspress.

## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -226,7 +226,7 @@ async function createInternalBuildConfig(
 
         chain.module
           .rule('css-virtual-module')
-          .test(/\.rspress\/runtime\/virtual-global-styles/)
+          .test(/\.rspress[\\/]runtime[\\/]virtual-global-styles/)
           .merge({ sideEffects: true });
       },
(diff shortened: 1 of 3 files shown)


packages/core/src/node/initRsbuild.ts
```diff
@@ -223,6 +223,11 @@ async function createInternalBuildConfig(
         }
 
         chain.resolve.extensions.prepend('.md').prepend('.mdx').prepend('.mjs');
+
+        chain.module
+          .rule('css-virtual-module')
+          .test(/\.rspress\/runtime\/virtual-global-styles/)
```
chenjiahan, line 229:
```suggestion
          .test(/\.rspress[\\/]runtime[\\/]virtual-global-styles/)
```

chenjiahan (approved):

PR #1091, opened 2024-05-17
fix: do not display no search result when search not completed

Timeless0911:
## Summary

do not display no search result when search not completed

## Checklist


Timeless0911:
!canary

Timeless0911:
!canary

PR #1104, opened 2024-05-21
feat: add search.loading options for remote search

Timeless0911:
## Summary

- add search.loading options for remote search
- remove  1s of search index loading for local search

## Checklist

### DIFF
--- a/packages/shared/src/types/index.ts
+++ b/packages/shared/src/types/index.ts
@@ -336,6 +336,7 @@ export type RemoteSearchOptions = SearchHooks & {
   domain?: string;
   indexName: string;
   searchIndexes?: RemoteSearchIndexInfo[];
+  searchLoading?: boolean;
 };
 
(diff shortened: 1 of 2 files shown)


Timeless0911:
!canary

Timeless0911:
!canary

Timeless0911:
!canary

packages/shared/src/types/index.ts
```diff
@@ -336,6 +336,7 @@ export type RemoteSearchOptions = SearchHooks & {
   domain?: string;
   indexName: string;
   searchIndexes?: RemoteSearchIndexInfo[];
+  searchLoading?: boolean;
```
chenjiahan, line 339: should we document this new option?

Timeless0911: remote search API is not stable and in long term will write a doc for it.

chenjiahan (approved):

PR #1110, opened 2024-05-23
fix(theme-default): sidebar scrollbar height error on ios safari and mask is invalid

### LINKED ISSUE #1103: [Bug]: sidebar scrollbar error and mask is invalid when using ios safari
### Version

```sh
System: iOS 
Browsers: Safari
```

### Details

When opening the sidebar, the document content can still be scrolled. In addition, there is a problem with the scrollbar area in the sidebar.
IMG_4991
IMG_4993

### Reproduce link

https://rspress.dev/

### Reproduce Steps

1. use ios safari
2. open [`rspress.dev`](https://rspress.dev)

shulaoda:
Fixed #1103 

## Related Issue
#1103 

## Checklist

### DIFF
--- a/packages/theme-default/src/components/Sidebar/index.module.scss
+++ b/packages/theme-default/src/components/Sidebar/index.module.scss
@@ -21,7 +21,7 @@
   background: var(--rp-c-bg);
 }
 
-.sidebar-content {
+.sidebarContent {
   overflow-y: auto;
   overflow-x: hidden;
   max-height: calc(100% - var(--rp-nav-height));
--- a/packages/theme-default/src/components/Sidebar/index.module.scss
+++ b/packages/theme-default/src/components/Sidebar/index.module.scss
@@ -21,14 +21,6 @@
   background: var(--rp-c-bg);
 }
 
-.sidebarContainer {
-  padding: 0 0 96px 0;
-}
-
-.sidebarContent {
-  overflow-x: hidden;
-}
-
 .sidebar.open {
   opacity: 1;
   visibility: visible;
@@ -59,10 +51,6 @@
 @media (min-width: 1440px) {
   .sidebar {
     width: var(--rp-sidebar-width);
-  }
-
-  .sidebarContainer {
-    padding-left: 0;
   }
 }
 
--- a/packages/theme-default/src/components/Sidebar/index.module.scss
+++ b/packages/theme-default/src/components/Sidebar/index.module.scss
@@ -21,6 +21,12 @@
   background: var(--rp-c-bg);
 }
 
+.sidebar-content {
+  overflow-y: auto;
+  overflow-x: hidden;
+  max-height: calc(100% - var(--rp-nav-height));
+}
+
 .sidebar.open {
   opacity: 1;
   visibility: visible;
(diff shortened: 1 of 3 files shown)


packages/theme-default/src/components/Sidebar/index.tsx
```diff
@@ -72,14 +77,26 @@ export let matchCache: WeakMap<
...
-  const [sidebarData, setSidebarData] = useState<
-    (ISidebarDivider | ISidebarItem | NormalizedSidebarGroup)[]
-  >(rawSidebarData.filter(Boolean).flat());
   const pathname = decodeURIComponent(rawPathname);
+  const langRoutePrefix = normalizeSlash(localesData.langRoutePrefix || '');
+
+  useEffect(() => {
+    if (inBrowser) {
+      if (isSidebarOpen) {
+        document.body.style.overflow = 'hidden';
```
Timeless0911, line 94: is this body style change have side effects

packages/theme-default/src/components/Sidebar/index.tsx
```diff
@@ -186,27 +203,24 @@ export function Sidebar(props: Props) {
...
-          </div>
+      {!uiSwitch.showNavbar ? null : (
+        <div className={styles.navTitleMask}>
+          <NavBarTitle />
         </div>
+      )}
+      <div
+        className="rspress-scrollbar"
+        style={{
+          maxHeight: 'calc(100% - var(--rp-nav-height))',
```
Timeless0911, line 214: use css module?

packages/theme-default/src/components/Sidebar/index.tsx
shulaoda: I referred to the implementation of `vitepress`, but I did not handle the special case where the overflow on the body may have an initial value.

Timeless0911: Every modification of the body style always brings some issues.... We need to be more cautious. Can you verify it more or can we not use the body style?

shulaoda: I think this is the best way, and its impact will only exist when the sidebar is opened.

shulaoda: There are some special circumstances that need to be considered and resolved.

packages/theme-default/src/components/Sidebar/index.module.scss
```diff
@@ -21,12 +21,10 @@
   background: var(--rp-c-bg);
 }
 
-.sidebarContainer {
-  padding: 0 0 96px 0;
-}
-
-.sidebarContent {
+.sidebar-content {
```
Timeless0911, line 24: camelcase is better to be consistent with other modules？

Timeless0911 (approved):

packages/theme-default/src/components/Sidebar/index.tsx
Timeless0911: nothing yet

chenjiahan:
Thanks! ❤️

PR #1129, opened 2024-05-29
fix: search item click target set

Timeless0911:
## Summary

we should use lodash.isEqual to judge whether two object-element array is same.

## Checklist


Timeless0911:
!canary

chenjiahan (approved):

Issue #1144, opened 2024-06-03
[Bug]: 最新版本内部使用 rsbuild@0.7.0 之后，自定义 lessOptions 无法生效


soulwu:
### 版本信息

```sh
System:
    OS: macOS 14.2.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 122.73 MB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 125.0.6422.113
    Safari: 17.2.1
  npmPackages:
    rspress: ^1.23.0 => 1.23.0
```

### 问题详情

rsbuild 从 0.7.0 版本开始不再内置 less-loader，改为使用 plugin 方式。rspress 新版本也跟进了这一改动，但是新的方式不再支持用户自定义配置 lessOptions。由于 rspress 内部默认加载了 @rsbuild/plugin-less，因此用户也无法自行加载该插件

### 复现链接

无

### 复现步骤

无法配置 lessOptions，比如 lessOptions: { javascriptEnabled: true, math: 'always' }

Timeless0911:
You can modify the `plugins` in `builderConfig`. The built-in plugins for less and sass will be removed in Rspress 2.0 in the future.

soulwu:
> You can modify the `plugins` in `builderConfig`. The built-in plugins for less and sass will be removed in Rspress 2.0 in the future.

https://codesandbox.io/p/devbox/bold-cdn-wpryxf?file=%2Fpackage.json%3A18%2C1

Timeless0911:
Rsbuild may can not modify plugins funtionally like `raw` or `tools.rspack` to remove the less plugins built-in and config new less plugins with some options.

You can config via `builderConfig.tools.rspack` to modify rspack config directly. 

Any other good idea? @chenjiahan

chenjiahan:
@Timeless0911 Rspress can check if user is configuring the Less/Sass plugin via `builderPlugins`. If configured, Rspress should not register these plugins again.

PR #1147, opened 2024-06-04
fix: support register pluginSass and pluginLess in config file

Timeless0911:
## Summary

closes: #1144 

## Checklist


packages/core/src/node/initRsbuild.ts
```diff
@@ -83,11 +87,23 @@ async function createInternalBuildConfig(
     node: ['node >= 14'],
   };
 
+  const sassPluginIncluded =
+    config.builderPlugins?.some(plugin => plugin.name === 'rsbuild:sass') ||
```
chenjiahan, line 91: We can import PLUGIN_SASS_NAME or PLUGIN_LESS_NAME from the plugins

packages/core/src/node/initRsbuild.ts
```diff
@@ -83,11 +87,23 @@ async function createInternalBuildConfig(
...
+
+  const lessPluginIncluded =
+    config.builderPlugins?.some(plugin => plugin.name === 'rsbuild:less') ||
+    config.builderConfig?.plugins?.some(
+      plugin => plugin && (plugin as RsbuildPlugin).name === 'rsbuild:less',
+    );
+
   return {
     plugins: [
       pluginReact(),
```
chenjiahan, line 104: This can be applied to pluginReact too, which can allow users to custom pluginReact options.

chenjiahan: @Timeless0911 cc

packages/core/src/node/initRsbuild.ts
chenjiahan, line 1: should we document this change?

Timeless0911: I prepare to mention it in `builderPlugins` field.

chenjiahan (approved):
玩

Issue #23, opened 2023-05-01
[Feature]: Doc supports client redirects


chenjiahan:
### What problem does this feature solve?

Support client-side redirects, like [@docusaurus/plugin-client-redirects](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-client-redirects).

The redirects can be used when we modify the URL address of some pages and need to keep the original URL accessible.

### What does the proposed API look like?

```js
// modern.config.ts
{
  doc: {
    redirects: [
      // /docs/oldDoc -> /docs/newDoc
      {
        to: '/docs/newDoc',
        from: '/docs/oldDoc',
      },
      // Redirect from multiple old paths to the new path
      {
        to: '/docs/newDoc2',
        from: ['/docs/oldDocFrom2019', '/docs/legacyDocFrom2016'],
      },
    ],
  }
}
```

PR #1166, opened 2024-06-11
feat: support client redirect

Timeless0911:
## Summary

closes: #23 

## Checklist


e2e/fixtures/client-redirects/doc/index.mdx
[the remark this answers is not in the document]
Timeless0911: Creating new HTML files for the former document before redirection requires traversing all existing routes. There are many boundary conditions for scenarios where routes and file directories are different (multiple versions, multiple languages) and regular expression matching. 

Furthermore, after moving a large number of file directories, generate additonal HTML files will cause the `doc_build` directory to be very messy.

After supporting the default generation of 404.html in #1171, we can rely on the deployment side (Nginx, netlify, github pages, etc...) to fallback to the 404 page, and thereby triggering the global redirect component.

At present, the implementation of this plugin seems not very urgent. If users want to implement redirection feature, they can also easily write a global component.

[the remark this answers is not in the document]
SoonIter: I think fallback to index.html rather than 404.html is better for consistent behavior between csr and ssr

https://github.com/web-infra-dev/rspack/blob/main/website/docs/public/_redirects

packages/document/docs/en/plugin/official-plugins/client-redirects.mdx
```diff
@@ -0,0 +1,119 @@
...
+
+Used for client redirects.
+
+::: warning NOTE
+
+For this plugin to work, you need to set the fallback related configuration to the `404.html` page in the build artifacts in the relevant deployment configuration.
+
+For example:
+
+```js title="nginx"
```
chenjiahan, line 15: It is not recommended to use code blocks in the warning container, the UI is really ugly 😂

Maybe we can just make it as a section, such as `## Precautions`

image

Timeless0911: done

packages/document/docs/zh/plugin/official-plugins/client-redirects.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+import { SourceCode, PackageManagerTabs } from 'rspress/theme';
+
+<SourceCode href="https://github.com/web-infra-dev/rspress/tree/main/packages/plugin-client-redirects" />
+
+用于客户端重定向。
+
+## 注意
+
+该插件的正常工作需要在相关部署配置中将 fallback 相关的配置设置为产物中的 `404.html` 页面。
```
chenjiahan, line 11:
```suggestion
为了使该插件正常工作，你需要在服务器的部署配置中，将 fallback 路径设置为 Rspress 生成的 `404.html` 页面。
```

packages/document/docs/en/plugin/official-plugins/client-redirects.mdx
```diff
@@ -0,0 +1,117 @@
...
+
+import { SourceCode, PackageManagerTabs } from 'rspress/theme';
+
+<SourceCode href="https://github.com/web-infra-dev/rspress/tree/main/packages/plugin-client-redirects" />
+
+Used for client redirects.
+
+## Precautions
+
+For this plugin to work correctly, you need to set the fallback related configuration to the `404.html` page in the build artifacts in the relevant deployment configuration.
```
chenjiahan, line 11:
```suggestion
To ensure that the plugin works correctly, you need to set the fallback path in the server's deployment configuration to the `404.html` page generated by Rspress.
```

chenjiahan (approved):

PR #1145, opened 2024-06-03
feat: support extra head tags

toteto:
## Summary

This PR adds support for extra head tags per page via Frontmatter config. 

## Related Issue

#1121 

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

PR #1168, opened 2024-06-11
feat: support extra head tags, extension to #1145

toteto:
## Summary

This PR is extension to #1145 to simplify the approval/merge process. 

The change is that it introduces additional types to the `head` config attribute. This addition enables better per-page headers, in a sense plugin header support.

The immediate benefit i see in addition to #1145 is less duplication of tags that are to be only for a subset of the pages.

## Related Issue

#1145

## Checklist


Timeless0911:
It seems the lockfile should be update to pass CI.

Timeless0911:
LGTM, I will translate the document into Chinese after this PR merged.

packages/shared/src/types/index.ts
```diff
@@ -274,6 +280,8 @@ export interface FrontMatterMeta {
   lineNumbers?: boolean;
   overviewHeaders?: number;
   titleSuffix?: string;
+  head?: [string, Record<string, string>][];
+  [key: string]: unknown;
```
Timeless0911, line 284: just for typing hint?

toteto: Because the pages frontmatter can include any key-value, this type allows reading of any additional properties/keys that have not been defined in the `FrontMatterMeta`. It also aids into making the typing for `loadFrontmatter` work with generic.

Timeless0911: sounds great

Timeless0911 (approved):

Timeless0911:
Thanks~

toteto:
> Thanks~

Thanks for getting support while doing this PR, much appreciate it

Issue #1158, opened 2024-06-06
[Feature]: Inject context in sidebar generate to enable using css to control sidebar display


Timeless0911:
### What problem does this feature solve?

In some huge scale sites, we may need to display different pages according to the switching of different modes, such as the react and vue modes of a tool library document. 

At this time, clicking the mode switching button is expected to cause the sidebar content to change.

### What does the proposed API look like?

It is hoped that `context` configuration will be supported in `_meta.jaon`, and by injecting `data-context` when generating the sidebar dom, users can control the display of sidebar via css when clicking the switch button, so as to achieve the purpose of controlling sidebar content display.

```json title="_meta.json"
{
  {
    "type": "file",
    "name": "foo.react",
    "context": "react"
  },
  {
    "type": "file",
    "name": "foo.vue",
    "context": "vue"
  }
}
```

```html
<div class=".rspress-sidebar-group">
  <div data-context="vue" />
  <div data-context="react" />
</div>
```

> This solution is a less invasive solution to Rspress and may cause other experience issues, such as the feature of go to  previous/next page

PR #1170, opened 2024-06-12
feat: support inject context when generate sidebar

Timeless0911:
## Summary

closes: #1158 

## Checklist


packages/document/docs/en/api/config/config-frontmatter.mdx
```diff
@@ -245,3 +245,9 @@ Or
 overviewHeaders: [2, 3]
 ---
 ```
+
+## context
+
+- Type: `string`
+
+After configuration, the `data-context` attribute will be added to the DOM node when the sidebar is generated, and the value is the configured value.
```
chenjiahan, line 253: Can we add the example from https://github.com/web-infra-dev/rspress/issues/1158 to our document?

This will make the usage more clear.

Timeless0911: added

chenjiahan (approved):

PR #1181, opened 2024-06-13
fix(theme-default): scroll restoration

### LINKED ISSUE #1154: [Bug]: Aside activate wrongly when refresh page which contain anchor hash in URL
### 版本信息

```sh
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 146.42 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 125.0.6422.142
    Safari: 17.5
  npmPackages:
    rspress: ^1.23.0 => 1.23.0
```

### 问题详情

image

After memorizing an anchor point, why doesn't it jump to the specific anchor point after refreshing, and why is the right side highlighted in the wrong place?

### 复现链接

https://rspress.dev/guide/advanced/custom-search#aftersearch

### 复现步骤

1. Click on one of the document's anchor points

2. The current url records the anchor.

3. Refresh the page according to the current url

4. Finds the problem mentioned above

shulaoda:
## Summary
The issue arises due to the browser's default behavior of `history.scrollRestoration`, which restores the scroll position upon refresh to enhance the user experience. 
I changed this default behavior in `theme-default` to avoid affecting other custom templates. However, if there is no anchor in the link, it affects the refresh experience to some extent. 
If we consider that it only works when there is an anchor, it will make the problem more complex.

## Related Issue
closed #1154 

## Checklist


Timeless0911:
If we don’t click the anchor so that the url does not contain `location.hash`(#abc), then the performance after refreshing the page will indeed damage the user’s reading experience.

Only works when there is an anchor is necessary I think. Just check whether the current pathname contains location.hash?

Timeless0911 (approved):

Timeless0911:
LGTM

PR #1195, opened 2024-06-20
fix: only append base to internal non-relative urls

Timeless0911:
## Summary

Only append base to internal non-relative urls, to prevent conflict between `base` and `output.assetPrefix` when using relative urls.

## Checklist

### DIFF
--- a/packages/runtime/src/utils.ts
+++ b/packages/runtime/src/utils.ts
@@ -46,6 +46,10 @@ export function normalizeImagePath(imagePath: string) {
   if (isExternalUrl(imagePath) || isDataUrl(imagePath)) {
     return imagePath;
   }
+  // only append base to internal non-relative urls
+  if (!imagePath.startsWith('/') || imagePath.startsWith('//')) {
+    return imagePath;
+  }
 
   return withBase(imagePath);
 }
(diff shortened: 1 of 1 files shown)


packages/runtime/src/utils.ts
```diff
@@ -46,6 +46,10 @@ export function normalizeImagePath(imagePath: string) {
   if (isExternalUrl(imagePath) || isDataUrl(imagePath)) {
```
SoonIter, line 46:
```suggestion
  if (isExternalUrl(imagePath) || isDataUrl(imagePath) || imagePath.startsWith('//')) {
```

packages/runtime/src/utils.ts
```diff
@@ -46,6 +46,10 @@ export function normalizeImagePath(imagePath: string) {
...
     return imagePath;
   }
+  // only append base to internal non-relative urls
+  if (!imagePath.startsWith('/') || imagePath.startsWith('//')) {
```
SoonIter, line 50:
```suggestion
  if (!imagePath.startsWith('/')) {
```

SoonIter (approved):

PR #1212, opened 2024-06-27
feat: better gracefully shutdown

Timeless0911:
## Summary

Actively exit the process when gracefully shutdown.

## Checklist

### DIFF
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -100,9 +100,14 @@ cli
       await startDevServer();
 
       const exitProcess = async () => {
-        await cliWatcher.close();
-        await devServer.close();
+        try {
+          await devServer.close();
+          await cliWatcher.close();
+        } finally {
+          process.exit(0);
+        }
       };
+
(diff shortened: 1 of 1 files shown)


packages/cli/src/index.ts
chenjiahan, line 1: The PR title can be `feat` instead of `chore`.

Timeless0911: done

chenjiahan (approved):

packages/cli/src/index.ts
```diff
@@ -100,9 +100,14 @@ cli
       await startDevServer();
 
       const exitProcess = async () => {
```
9aoy, line 102: shall we add a timeout to force exit after 5 seconds?

chenjiahan: 5 seconds too too long

Timeless0911: Welcome your pr 😀

Issue #1258, opened 2024-07-16
[Bug]: Aside component display not as expected when using \<xxx\> in md header


Timeless0911:
### Details

![image](https://github.com/user-attachments/assets/e9242313-b076-4d49-9051-5a30094208c0)

![image](https://github.com/user-attachments/assets/dbee7905-c1a7-407a-9998-2b4f2676dd54)

The opening parenthesis displays not as expected in the right navigation aside bar.

PR #1264, opened 2024-07-16
fix: replace < in render inline markdown

Timeless0911:
## Summary

closes: #1258 

replace `<list>` to prevent disappearing in dom, but not replace \<number\>

**source code**

image

**origin**
image

**now**
image

## Checklist


chenjiahan (approved):
👍

PR #1294, opened 2024-08-01
feat(search): upgrade flexsearch 0.7 and adopt the new API design

zoolsher:
## Summary

flexsearch 0.7.x designed new API for document search this commit adopted the new API

## Checklist

### DIFF
--- a/packages/plugin-api-docgen/package.json
+++ b/packages/plugin-api-docgen/package.json
@@ -45,7 +45,7 @@
   "peerDependencies": {
     "react": ">=17.0.0",
     "react-router-dom": "^6.8.1",
-    "@rspress/core": "workspace:*",
+    "@rspress/core": "workspace:^1.0.2",
     "typescript": "^5.5.3"
   },
   "peerDependenciesMeta": {
--- a/packages/plugin-api-docgen/package.json
+++ b/packages/plugin-api-docgen/package.json
@@ -45,7 +45,7 @@
   "peerDependencies": {
     "react": ">=17.0.0",
     "react-router-dom": "^6.8.1",
-    "@rspress/core": "^1.0.2",
+    "@rspress/core": "workspace:*",
     "typescript": "^5.5.3"
   },
   "peerDependenciesMeta": {
--- a/packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
+++ b/packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
@@ -15,7 +15,7 @@ import type { SearchOptions } from '../types';
 import { LOCAL_INDEX, type Provider, type SearchQuery } from '../Provider';
 import { normalizeTextCase } from '../util';
 
-type TFlexSearchDocumentWithType = Document<PageIndexInfo, true>;
+type FlexSearchDocumentWithType = Document<PageIndexInfo, true>;
 
 interface PageIndexForFlexSearch extends PageIndexInfo {
   normalizedContent: string;
--- a/packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
+++ b/packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
@@ -40,11 +40,11 @@ function tokenize(str: string, regex) {
 }
 
 export class LocalProvider implements Provider {
-  #index?: TFlexSearchDocumentWithType;
+  #index?: FlexSearchDocumentWithType;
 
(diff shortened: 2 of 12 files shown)


packages/plugin-api-docgen/package.json
```diff
@@ -45,7 +45,7 @@
   "peerDependencies": {
     "react": ">=17.0.0",
     "react-router-dom": "^6.8.1",
-    "@rspress/core": "^1.0.2",
+    "@rspress/core": "workspace:*",
```
chenjiahan, line 48: Using `workspace:*` will generate a fixed version number, which forces the plugin to use the same version as `@rspress/core`.

It is usually more appropriate to define a version range in peerDependencies, such as `workspace:^1.26.3` (Modern.js uses this format), and changeset will bump the version in `workspace:^1.26.3`.

packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
```diff
@@ -3,14 +3,20 @@
...
+  IndexOptionsForDocumentSearch,
+} from 'flexsearch';
+import FlexSearchDocument from 'flexsearch/dist/module/document';
 import searchIndexHash from 'virtual-search-index-hash';
 import { type PageIndexInfo, SEARCH_INDEX_NAME } from '@rspress/shared';
 import type { SearchOptions } from '../types';
 import { LOCAL_INDEX, type Provider, type SearchQuery } from '../Provider';
 import { normalizeTextCase } from '../util';
 
+type TFlexSearchDocumentWithType = Document<PageIndexInfo, true>;
```
chenjiahan, line 18: The `T` prefix can be omitted, we have no convention to add a `T` prefix 😄

```suggestion
type FlexSearchDocumentWithType = Document<PageIndexInfo, true>;
```

packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
```diff
@@ -63,61 +69,83 @@ export class LocalProvider implements Provider {
...
+      {
+        tokenize: 'full',
+        document: {
+          id: 'id',
+          store: true,
+          index: ['normalizedTitle', 'headers', 'normalizedContent'],
+        },
+        cache: 100,
+        // charset: {
+        //   split: /\W+/,
```
chenjiahan, line 82: It this config no longer required?

packages/plugin-api-docgen/package.json
zoolsher: Isn't a fixed version better? Since we don't have a CI process to guarantee that the plugin can work with another version of `@rspress/core`.

packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
zoolsher: They made a breaking change to this API; I'm not sure about the effect for now. So, I saved the configuration in a comment, maybe someday we will need to revert to it.

chenjiahan: get 👌

packages/plugin-api-docgen/package.json
chenjiahan: Fixed version is too strict, in most cases Rspress plugins can work with future versions of `@rspress/core`. As we do not remove any plugin API in non-major releases.

Most plugin systems in the community choose a loose peer dependency declaration, usually declaring one or more major versions.

Examples:

- Vite: https://github.com/vitejs/vite-plugin-react/blob/main/packages/plugin-react/package.json#L47-L49
- webpack plugin: https://github.com/jantimon/html-webpack-plugin/blob/main/package.json#L72-L75
- webpack loader: https://github.com/webpack-contrib/sass-loader/blob/master/package.json#L44-L50
- babel: https://github.com/babel/babel/blob/main/packages/babel-plugin-syntax-typescript/package.json#L23-L25

Timeless0911: Changeset will bump a major version for plugin after we add a minor or major changset for core if we use a fixed version in a peerdeps.

zoolsher: OK I'll use `workspace:^`.

zoolsher: updated

chenjiahan (dismissed):

chenjiahan (approved):

PR #1312, opened 2024-08-08
fix(watch): improve watch solution, ignore _meta.json out of doc root dir

zllkjc:
## Summary

before:
1. rspress will watch all file in doc root dir, but only handle some config file.
2. all the `_meta.json` in cwd are watched.

now
1. only watch `_meta.json` files in doc root dir.

## Checklist


Timeless0911 (approved):
🤙

Issue #1317, opened 2024-08-08
[Feature]: support don't fallback to csr when ssr failed


hardfist:
### What problem does this feature solve?

sometime ssr failed means there's something wrong with the code, if we ignore the ssg failed and fallback to csr which may cause csr fail too, which happens in https://github.com/web-infra-dev/rspack/issues/7505

### What does the proposed API look like?

add a option to disable ssg fallback 
```ts
{
  fallback: false | 'csr' // false mean don't do fallback when ssg failed , 'csr' means fallback to csr when ssg failed
}
```

zoolsher:
I believe I understand the context. The issue seems to be not with the fallback, but with what happens when SSG fails. You want the CI to break if the document's SSG fails, correct? Maybe we could provide a configuration option like `{ ssg: { strict: true } }`.

hardfist:
I discussed with @chenjiahan before and he thinks we may have different fallback strategy(csr or others), so fallback is more extensible than strict

zoolsher:
Yes, but this seems more like a behavior of SSG rendering that determines whether it breaks the CI. And I think `strict: true` is more likely to break your CI pipeline than `fallback: false`.

chenjiahan:
Rspress v1 enables the CSR fallback by default, and we can not change the default behavior in non-major releases. The default behavior will be changed in Rspress v2.

So we plan to add the new `fallback` config first, and in Rspress v2, the default value of `fallback` will be `false`.

When users want to use the CSR fallback, they can add `fallback: 'csr'` to their config file. There can be more fallback types in the future.

chenjiahan:
My bad, I didn't clearly explain the context of this option.

SSG will be `strict` by default in Rspress v2. And we will provide a config to allow users to opt-in to the current fallback behavior.

In this case, `ssg.fallback` is more semantic because it explicitly tells what the option does.

- `ssg.fallback`:

```js
export default {
  ssg: {
    fallback: 'csr',
  }
};
```

- `ssg.strict`:

```js
export default {
  ssg: {
    strict: false,
  }
};
```

zoolsher:
> SSG will be strict by default in Rspress v2. 

by default or by force?

chenjiahan:
by default. If users choose to use fallback, it is no longer strict.

zoolsher:
Still, `fallback: false` doesn’t seem like a configuration that would break the build or throw an error.

[...]

PR #1325, opened 2024-08-13
feat: add new `ssg.strict` config

chenjiahan:
## Summary

Add a new `ssg.strict` config.

```js
export default {
  ssg: {
    strict: true,
  },
};
```

## Related Issue

See https://github.com/web-infra-dev/rspress/issues/1317 for more background.

## Checklist


Timeless0911 (approved):

packages/core/src/node/build.ts
```diff
@@ -80,18 +82,25 @@ export async function renderPages(
     // 2. ssr bundle render failed
     // 3. ssg is disabled
     let render = null;
-    if (enableSSG) {
+    if (ssgConfig) {
       try {
         const { default: ssrExports } = await import(
           pathToFileURL(ssrBundlePath).toString()
         );
         ({ render } = ssrExports as SSRBundleExports);
```
zoolsher, line 90: This render function gets invoked later to render the page dom. There is also a try catch. Should we also add a throw error there?

chenjiahan: Nice catch, I missed that.

Updated 😄

packages/core/src/node/build.ts
```diff
@@ -129,10 +138,14 @@ export async function renderPages(
             try {
               ({ appHtml } = await render(routePath, helmetContext.context));
             } catch (e) {
+              if (typeof ssgConfig === 'object' && ssgConfig.strict) {
```
zoolsher, line 141: Do you think we should provide additional context to assist users with debugging? like add `routePath` or more to the error.message

chenjiahan: Sure 👍  the error logs has been updated:

Screenshot 2024-08-14 at 11 48 44

zoolsher: Cool. Really helpful error diagnostic

zoolsher (approved):
LGTM

PR #1329, opened 2024-08-14
fix: window unexpected href when feature link is not configured

Timeless0911:
## Summary

`normalizeHrefInRuntime(withBase(rawLink, routePath))` will generate a `/index`, which will cause unexpected window href if user click a feature component without link configured.

## Checklist


packages/theme-default/src/components/HomeFeature/index.tsx
chenjiahan, line 1: can we add a test case

Timeless0911: done

chenjiahan (approved):

Issue #1344, opened 2024-08-21
[Feature]: Is is better to use __webpack_public_path__ for search index url? 


liuwenzhuang:
### What problem does this feature solve?

My site static files is deploy to oss, and html files is render from a node.js server, so my config is like:

```ts
builderConfig: {
    output: {
      assetPrefix: `{{STATIC_PATH}}/`,
    }
}
```

and node.js server will replace `{{STATIC_PATH}}` placeholder to oss url when render, and I override `__webpack_public_path__` to ensure dynamic files path is correct.

After deploy, the page's resource is ok, but the search json file's request url is like `http://mysite.com/{{STATIC_PATH}}/static/search_index.json`, not prefix with `__webpack_public_path__`.

### What does the proposed API look like?

Maybe the `__webpack_public_path__` prefix or `assetPrefix` prefix to search index url could be config?

Timeless0911:
How you override `__webpack_public_path__`

chenjiahan:
@Timeless0911 `__webpack_public_path__` can be overriden in the runtime code.

https://rspack.dev/api/runtime-api/module-variables#__webpack_public_path__

PR #1347, opened 2024-08-21
fix: use `__webpack_public_path__` for search index url

Timeless0911:
## Summary

closes: #1344 

## Checklist

### DIFF
--- a/packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
+++ b/packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
@@ -52,7 +52,7 @@ export class LocalProvider implements Provider {
     const searchIndexLang = lang ? `.${lang}` : '';
 
     const result = await fetch(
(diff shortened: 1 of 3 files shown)


packages/theme-default/src/components/Search/logic/providers/LocalProvider.ts
```diff
@@ -52,7 +52,7 @@ export class LocalProvider implements Provider {
     const searchIndexLang = lang ? `.${lang}` : '';
 
     const result = await fetch(
-      `${process.env.__ASSET_PREFIX__}/static/${SEARCH_INDEX_NAME}${searchIndexVersion}${searchIndexLang}.${searchIndexHash[searchIndexGroupID]}.json`,
+      `${__webpack_public_path__}static/${SEARCH_INDEX_NAME}${searchIndexVersion}${searchIndexLang}.${searchIndexHash[searchIndexGroupID]}.json`,
```
chenjiahan, line 55: Should we check if `__webpack_public_path__` ends with a slash?

Timeless0911: updated

Issue #1357, opened 2024-08-28
[Bug]: yarn create rspress@latest Failed


Alkaidcc:
### Version

```sh
System:
    OS: macOS 14.6.1
    CPU: (8) arm64 Apple M2
    Memory: 532.83 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 127.0.6533.122
    Safari: 17.6
```

### Details

yarn create rspress@latest
yarn create v1.22.22
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...

success Installed "create-rspress@1.28.2" with binaries:
      - create-rspress
/bin/sh: /Users/.yarn/bin/create-rspress@latest: No such file or directory
error Command failed.
Exit code: 127
Command: /Users/.yarn/bin/create-rspress@latest
Arguments:
Directory: /Users/playground
Output:

info Visit https://yarnpkg.com/en/docs/cli/create for documentation about this command.

yarn version: 1.22.22

### Reproduce link

https://rspress.dev/zh/guide/start/getting-started

### Reproduce Steps

1. run `yarn create rspress@latest`

Timeless0911:
You can use `yarn create rspress` for yarn especially, I will update the docs soon, thank you!

PR #1358, opened 2024-08-28
fix: yarn create command in PackageManagerTabs

Timeless0911:
## Summary

Do special treatment for `yarn create` command, do not use `@latest`

#1357 

## Checklist

### DIFF
--- a/packages/theme-default/src/components/PackageManagerTabs/index.tsx
+++ b/packages/theme-default/src/components/PackageManagerTabs/index.tsx
@@ -23,9 +23,15 @@ export interface PackageManagerTabProps {
 }
 
 function normalizeCommand(command: string): string {
+  // If command is yarn create foo@latest, remove `@latest`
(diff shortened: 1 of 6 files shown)


packages/theme-default/src/components/PackageManagerTabs/index.tsx
```diff
@@ -23,9 +23,15 @@ export interface PackageManagerTabProps {
 }
 
 function normalizeCommand(command: string): string {
+  // If command is yarn create foo@latest, remove `@latest`
+  if (command.startsWith('yarn create')) {
+    return command.replace(/(yarn create [^\s]+)@latest/, '$1');
```
chenjiahan, line 28: can we add a test case for this

Timeless0911: done

chenjiahan (approved):

PR #1374, opened 2024-09-01
feat: route README.md => index

ruochuan12:
## Summary

feat: route README.md => index

links: [vuepress route README.md => index](https://vuepress.github.io/zh/guide/page.html#%E8%B7%AF%E7%94%B1)

## Checklist

### DIFF
--- a/packages/core/src/node/route/RouteService.ts
+++ b/packages/core/src/node/route/RouteService.ts
@@ -71,6 +71,7 @@ export const normalizeRoutePath = (
     // remove the extension
     .replace(new RegExp(`\\.(${extensions.join('|')})$`), '')
     .replace(/\.html$/, '')
+    .replace(/\/readme$/i, '/')
     .replace(/\/index$/, '/');
 
   // restore the trail slash
--- a/packages/document/docs/en/guide/basic/conventional-route.mdx
+++ b/packages/document/docs/en/guide/basic/conventional-route.mdx
@@ -23,6 +23,7 @@ The specific mapping rules are as follows:
 
 | file path       | route path |
 | --------------- | ---------- |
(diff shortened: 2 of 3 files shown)


Timeless0911:
Thank you for your submission! Can you sign CLA to let CI pass?

Timeless0911:
I think we should document how to handle conflicts with `index.md` and add an e2e test case to cover this feature.

Timeless0911:
Hi @ruochuan12 , there are some points to discuss with you about this feature:
1. We do not want to introduce this default behavior to affect the original users since the default route of `readme.md` is not a standardized behavior and many other frameworks do not do this, like Vitepress and Docusaurus.
2. A more appropriate approach is to enable this behavior through a configuration,  which is not a easy work, changes in PR should be far from enough. We need complete tests to ensure that route generation, sidebar generation, link href, etc. work properly in all scenes like all files use index / all files use readme / files use mixed readme&index. And how to deal the conflict of two files should be mainly concerned.

ruochuan12:
Okay, let's not deal with it for now. I look forward to having such a feature in the future.

packages/core/src/node/route/RouteService.ts
```diff
@@ -71,6 +71,7 @@ export const normalizeRoutePath = (
     // remove the extension
     .replace(new RegExp(`\\.(${extensions.join('|')})$`), '')
     .replace(/\.html$/, '')
+    .replace(/\/readme$/i, '/')
```
JounQin, line 74: What if there are both `index.md` and `README.md`?

SoonIter:
closed 

use `index.md` is enough for now

It will increase the complexity of the routing, if someone needs this pr, they can open a new issue.

PR #1376, opened 2024-09-02
fix(core): fix writeSearchIndex function use userConfig instead modifiedConfig

liuwenzhuang:
problem: dev command make search_index.json write to default outDir when outDir is set in plugin, when search trigger, dev server will halt because search_index.json not exists in final outDir.

## Checklist


Timeless0911 (approved):
LGTM, thank you!

PR #1391, opened 2024-09-05
fix: remove margin-left for single a tag in SocialLinks component

huyixi:
This commit fixes an issue where a single <a> tag in the SocialLinks component would have an unwanted margin-left, causing misalignment. The issue is resolved by using the :only-child pseudo-class to remove the margin-left when there is
only one <a> tag.

Updated index.module.scss to include the necessary CSS selector.

### DIFF
--- a/packages/theme-default/src/components/SocialLinks/index.module.scss
+++ b/packages/theme-default/src/components/SocialLinks/index.module.scss
@@ -12,6 +12,10 @@
   }
 }
 
+a:only-child .social-links-icon {
+  margin-left: 0;
+}
+
 @media (max-width: 768px) {
   .menu-item:before {
     display: none;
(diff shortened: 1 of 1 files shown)


Timeless0911:
## before
image


## after
image

You can see the display. Your PR makes the left and right spacing not look well when there is only one social link.

image

The margins to the language switch button are too small.

huyixi:
> ## before
> image
> 
> ## after
> image
> 
> You can see the display. Your PR makes the left and right spacing not look well when there is only one social link.
> 
> image
> 
> The margins to the theme switch button are too small.

I forgot to consider the situation on large screens. The new PR has fixed this issue

Timeless0911 (approved):

PR #1451, opened 2024-10-08
feat(theme-default): children, outline and fixes for Badge

Huxpro:
## Summary

> Shout out to shadcn/ui and Vitepress for inspiration.

### Improve built-in `<Badge>` components:
- Custom children: this is particularly useful for adding icons.
- `outline` support: more variations!
- Style tweak:
  - use CSS variables from containers to look consistent.
  - fixes default badge being invisible in Dark mode.
  - better vertically aligned with different headings.
- Changes to `type` props:
  - add `info` type to be consistent with containers.
  - `tip` is added as the new default.

Do note that even though this kept untyped `<Badge>` unchanged, it
does introduce a **benign** breaking change to `<Badge type="info">`.

### Before vs. After

Light mode:
  
image

Dark mode:

image

### Changes to Docs:
- Badges docs are updated to reflect the new usages and new types.
- This MR correct "Internal Components" to be "Built-in Components".
      (does not involve URL changes due to the use of fragments)

## Related Issue

Dunno.

## Checklist

### DIFF
--- a/packages/document/docs/en/fragments/builtin-components.mdx
+++ b/packages/document/docs/en/fragments/builtin-components.mdx
@@ -6,7 +6,19 @@ The Badge component is used to display a badge. For example:
 import { Badge } from '@theme';
 
 function App() {
+  // Using text prop
   return <Badge text="info" type="info" />;
+
+  // Using custom children
+  return (
+    <Badge>
+      <img
+        style={{ height: '18px' }}
+        src="https://lf3-static.bytednsdoc.com/obj/eden-cn/uhbfnupenuhf/rspress/rspress-logo.png"
+      />
+      <span>Rspress</span>
+    </Badge>
+  );
 }
 ```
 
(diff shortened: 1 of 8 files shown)


packages/document/docs/en/fragments/builtin-components.mdx
```diff
@@ -6,26 +6,79 @@ The Badge component is used to display a badge. For example:
...
+<Badge type="info">
+  <SearchSvg width="16" />
+  <span>Github</span>
+</Badge>
+
+Inlined with text <Badge text="Tip" />
+
+##### H5 <Badge text="Info" type="info" />
+
+#### H4 <Badge text="Warning" type="warning" />
```
Timeless0911, line 56: We may only need H5 for `Inlined with Heading` example, since H4 and H3 will display in Aside component at right side. 
image

packages/document/docs/zh/fragments/builtin-components.mdx
```diff
@@ -6,26 +6,79 @@ Badge 组件用于展示状态的标记。使用方法如下：
...
+<Badge type="info">
+  <SearchSvg width="16" />
+  <span>Github</span>
+</Badge>
+
+内联文本 <Badge text="Tip" />
+
+##### H5 <Badge text="Info" type="info" />
+
+#### H4 <Badge text="Warning" type="warning" />
```
Timeless0911, line 56: Same as above

packages/document/docs/zh/fragments/builtin-components.mdx
```diff
@@ -6,26 +6,79 @@ Badge 组件用于展示状态的标记。使用方法如下：
...
+
+### H3 <Badge text="Danger" type="danger" />
 
 其中包含的 props 类型如下：
 
 ```ts
 interface BadgeProps {
```
Timeless0911, line 63: We should use Chinese here

Timeless0911:
Very detailed PR and documentation description, thank you 👍

packages/document/docs/en/fragments/builtin-components.mdx
Huxpro: @Timeless0911 How is this a problem? I don’t want to mistake purpose for appearance—the purpose is to faithfully demonstrate that the Badge component can accommodate different levels of headings.

Timeless0911: Got it, resolved it by no change.

Timeless0911 (approved):

chenjiahan:
Awesome 👍

Issue #1457, opened 2024-10-09
[Bug]: yarn dev pending when using some special characters as title


killdada:
### 版本信息

  System:
    OS: Windows 10 10.0.19041
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1230 v5 @ 3.40GHz
    Memory: 17.40 GB / 31.95 GB
  Browsers:
    Edge: Spartan (44.19041.1.0)
    Internet Explorer: 11.0.19041.1
  npmPackages:
    rspress: ^1.32.0 => 1.32.0

### 问题详情

![Image](https://github.com/user-attachments/assets/84fa1770-1ac1-4670-a88a-a88d03819198)

### 复现链接

https://codesandbox.io/p/devbox/goofy-yalow-94jm3q?workspaceId=efd5821f-9fb5-42c6-81b3-e2f553205837

### 复现步骤

yarn dev 卡住

Timeless0911:
https://github.com/web-infra-dev/rspress/blob/e2ebf7a26cb19f5f8230c39b09aa62f7336b5752/packages/core/src/node/runtimeModule/siteData/extractPageData.ts#L128-L141

This is related to here, I will fixed soon this week.

PR #1459, opened 2024-10-10
fix: unexpect pending when title include -number

9aoy:
## Summary

fix https://github.com/web-infra-dev/rspress/issues/1457
## Checklist


e2e/fixtures/basic/doc/index.mdx
```diff
@@ -1,5 +1,7 @@
 # Hello World
 
+### -22222222
```
chenjiahan, line 3: It is a bit weird to include this case in the `basic` fixture. We can create another fixture for this case.

Timeless0911: And we should using multiple `-1111111111` to test, thus TOC will generate with `-1111111111` and `-1111111111-1`

```
## -1111111111

## -1111111111
```

Timeless0911 (approved):

PR #1481, opened 2024-10-15
fix: load i18n source without cache

### LINKED ISSUE #1480: [Bug]: `i18nSourcePath` are cached via `require` which results restarting dev server does not take effect
### 版本信息

  System:
    OS: Linux 6.1 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (2) x64 AMD EPYC
    Memory: 2.40 GB / 4.01 GB
    Container: Yes
    Shell: Unknown
  npmPackages:
    rspress: ^1.33.1 => 1.33.1

### 问题详情

https://github.com/web-infra-dev/rspress/blob/bc68277ca60273e4496a10b8b41dead18a432395/packages/core/src/node/runtimeModule/i18n.ts#L11

Image

### 复现链接

https://codesandbox.io/p/devbox/fervent-solomon-54x6ld?file=%2Fi18n.json%3A5%2C2

### 复现步骤

1. `npm run dev`
2. change `i18n.json`

The dev server will restart and page will reload, but `i18n.json` changes do not take effect

---

`delete require.cache[i18nSourcePath]` before `require(i18nSourcePath)` should do the trick.

JounQin:
## Related Issue

close #1480

## Checklist

### DIFF
--- a/packages/core/src/node/runtimeModule/i18n.ts
+++ b/packages/core/src/node/runtimeModule/i18n.ts
@@ -7,6 +7,7 @@ const DEFAULT_I18N_SOURCE = join(process.cwd(), 'i18n.json');
 export function getI18nData(docConfig: UserConfig) {
   const { i18nSourcePath = DEFAULT_I18N_SOURCE } = docConfig;
   try {
+    delete require.cache[i18nSourcePath];
     // eslint-disable-next-line import/no-dynamic-require
     const i18nSource = require(i18nSourcePath);
     return i18nSource;


Timeless0911:
Can we directly readjson instead? I'm not quite sure about the performance difference between the two.

https://github.com/web-infra-dev/rspress/blob/bc68277ca60273e4496a10b8b41dead18a432395/packages/plugin-auto-nav-sidebar/src/walk.ts#L39-L40

Timeless0911 (approved):

PR #1507, opened 2024-10-20
fix: correctly iterate over plugins when using `removePlugin`

### LINKED ISSUE #1487: [Bug]: will not traverse all plugins if a plugin uses remove `removePlugin` or `addPlugin` API
### Details

For example:

```ts
plugins: [
  pluginPreview(),
  pluginPlayground(),
  pluginTypeDoc(),
],
```

Since `@rspress/plugin-playground` remove `@rspress/plugin-preview` in `config` function

https://github.com/web-infra-dev/rspress/blob/e4711144a19a710f910741e0ea73e0367db7f503/packages/plugin-playground/src/cli/index.ts#L69-L75

When pluginDriver run `config` function of each plugin, the `config` fuction of `@rspress/plugin-typedoc` which registered after `@rspress/plugin-playground` will not be executed.

This is because it uses`this.#plugins` to iterate.

https://github.com/web-infra-dev/rspress/blob/e4711144a19a710f910741e0ea73e0367db7f503/packages/core/src/node/PluginDriver.ts#L111-L124

Yukiniro:
## Summary

When calling modifyConfig, if some plugins call `removePlugin` or `addPlugin`, the index of the plugin in `this.#plugins` will change. Since `addPlugin` uses `push`, it will not affect the index of other plugins. If it is `removePlugin`, it is necessary to consider whether the plugin to be removed is before the current traversal index. If so, the plugins after the index will move forward 1 in `this.#plugins`. If not, there will be no impact.

## Related Issue

Closes #1487

## Checklist

### DIFF
--- a/packages/core/src/node/PluginDriver.ts
+++ b/packages/core/src/node/PluginDriver.ts
@@ -108,13 +108,22 @@ export class PluginDriver {
   async modifyConfig() {
     let config = this.#config;
 
-    for (const plugin of this.#plugins) {
+    for (let i = 0; i < this.#plugins.length; i++) {
+      const plugin = this.#plugins[i];
       if (typeof plugin.config === 'function') {
         config = await plugin.config(
           config || {},
           {
             addPlugin: this.addPlugin.bind(this),
-            removePlugin: this.removePlugin.bind(this),
+            removePlugin: (pluginName: string) => {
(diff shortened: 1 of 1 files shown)


Timeless0911 (approved):
LGTM

PR #1539, opened 2024-10-30
fix: langRoutePrefix should end with slash

JounQin:
## Summary

`langRoutePrefix` is used at https://github.com/web-infra-dev/rspress/blob/ae1bf881bbed54b13c7f31d457eb9984806c50d3/packages/theme-default/src/components/Nav/NavBarTitle.tsx#L47

Without the trailing slash, for located pages, it renders `/zh` for example, which means `/zh.html` but it should be `/zh/index.html` actually. And this change matches the default `langRoutePrefix` (`/`).

## Checklist

### DIFF
--- a/packages/theme-default/src/logic/useLocaleSiteData.ts
+++ b/packages/theme-default/src/logic/useLocaleSiteData.ts
@@ -1,4 +1,4 @@
-import type { NormalizedLocales } from '@rspress/shared';
+import { addTrailingSlash, type NormalizedLocales } from '@rspress/shared';
 import { usePageData } from '@rspress/runtime';
 
 export function useLocaleSiteData(): NormalizedLocales {
@@ -26,6 +26,6 @@ export function useLocaleSiteData(): NormalizedLocales {
 
   return {
     ...localeInfo,
-    langRoutePrefix: lang === defaultLang ? '/' : lang,
+    langRoutePrefix: lang === defaultLang ? '/' : addTrailingSlash(lang),
   } as NormalizedLocales;
 }


Timeless0911:
I think `/zh.html` can be hreffed correctly, there is no need to change the public hook of `useLocaleSiteData`

JounQin:
@Timeless0911 The default `langRoutePrefix` is `/`, so I believe `langRoutePrefix` should always end with slash?

Timeless0911:
> The default `langRoutePrefix` is `/`, so I believe `langRoutePrefix` should always end with slash?

I cannot confirm whether the modification of this hook will affect the use in other places and other users who import it at runtime. In addition, personally I think `langRoutePrefix ` should be  `''` and `zh`, just like what sidebar component do in 

https://github.com/web-infra-dev/rspress/blob/ae1bf881bbed54b13c7f31d457eb9984806c50d3/packages/theme-default/src/components/Sidebar/index.tsx#L93

This should be optimized with `normalizeHref ` function together in the future, `rspress.dev/index` `rspress.dev/zh/index` is ugly than `rspress.dev` `rspress.dev/zh`
https://github.com/web-infra-dev/rspress/blob/ae1bf881bbed54b13c7f31d457eb9984806c50d3/packages/shared/src/runtime-utils/index.ts#L235-L263

JounQin:
> In addition, personally I think `langRoutePrefix`  should be `''` and `zh`, just like what sidebar component do in

There is a problem that `/zh/` (`/zh/index.html`) and `/zh` (`/zh.html`) is totally different.

When toggling language, it goes to `/zh/index` in https://github.com/web-infra-dev/rspress/blob/main/packages/theme-default/src/components/Nav/NavVersions.tsx.

Timeless0911:
This should be an issue related to the href link issue of home page in different languags. 

## Current

- navbarTitle zh
image

- navbarTitle en
image

- translation menu zh
image

- translation menu en
image


## Expected

- `rspress.dev` in en (default lang)
- `rspress.dev/zh` in zh

Timeless0911:
> There is a problem that `/zh/` (`/zh/index.html`) and `/zh` (`/zh.html`) are totally different.

Yes, this is a question that needs to be considered. The situation is different when there is an html suffix or not (`cleanUrls` enabled or not).

JounQin:
> `rspress.dev/zh` in zh

Is this really expected? As I mentioned, `/zh` means `/zh.html` which SSG does not generates at all, `rspress.dev/zh/` is fine for `/zh/index.html`.

Timeless0911:
> Is this really expected? As I mentioned, `/zh` means `/zh.html` which SSG does not generates at all, `rspress.dev/zh/` is fine for `/zh/index.html`.

I know what you mean, we should ensure the request is `https://rspress.dev/zh/`.

Timeless0911:
In other words, your changes are no problem for resource requests, but the href display will appear a bit bloated. My goal is to simplify the content after the mouse hovers over it.

image

Timeless0911 (approved):

Timeless0911:
I think this PR can be merged for better semantics of link.

PR #1556, opened 2024-11-05
feat: display title in frontmatter as heading fallback

### LINKED ISSUE #1440: [Feature]: Use font awesome `title: <title>` as H1 heading if no H1 heading is added

Given following example `index.md` file:

```md
---
title: Heading 1
---

## Heading 2

Some content
```

In Docusaurus this would add `Heading 1` as the title of the document. But in RSPress:

Image

One can add an additional `Heading 1` in the document, but then markdownlint will complain about the document having two top-level headings, MD025.

* <https://github.com/DavidAnson/markdownlint/blob/main/doc/md025.md>

Image

I have no idea.

JounQin:
## Related Issue

close #1440

## Checklist


Timeless0911:
Thanks, I think this should be an opt-in feature with documented.

chenjiahan:
@Timeless0911 This feature should be the default behavior as it does not introduce breaking changes. Do you have any concerns?

Timeless0911:
> @Timeless0911 This feature should be the default behavior as it does not introduce breaking changes. Do you have any concerns?

The behavior of automatically adding titles should be made optional with `exclude` options. Users may insert relevant titles by themselves through other slots or scripts, and we need some ways to disable this behavior on some pages such as overview page.

chenjiahan:
> Users may insert relevant titles by themselves through other slots or scripts,

This doesn't seem to be common. It's relatively rare for the page title to be different from the h1 heading. I prefer providing an option to opt-out rather than opt-in.

Timeless0911:
> This doesn't seem to be common. It's relatively rare for the page title to be different from the h1 heading. I prefer providing an option to opt-out rather than opt-in.

Agreed. Providing an options is needed, and the default value can be true though.📌

JounQin:
I added an `fallbackHeadingTitle` themeConfig option.

Timeless0911 (approved):
LGTM

JounQin:
> I remember that in overview component, we deal title in a specific way, can you compatible with this case?

The `title` already reads `frontmatter.title`. I'm not sure what you expect to render.

Current:

```mdx
---
overview: true
title: API Overview
---

# xxxx
```

image

```mdx
---
overview: true
---

# xxxx
```

image

Timeless0911:
In this case, I think we should avoid to render two H1 in content?

JounQin:
> In this case, I think we should avoid to render two H1 in content?

I believe this is out of scope of this PR, the behavior doesn't change between it, so maybe a new PR for it is more suitable.

> And what will be rendered if we write like below that without H1 xxxx

It doesn't change:

image

Timeless0911:
Oh I forget that `Overview` do not use `docContent` to render. 

I think this PR is good enough to merge, thank you!

Issue #1536, opened 2024-10-29
[Bug]: Title with "\<" (escaped brackets) will crush the API overview


Huxpro:
### Version

  System:
    OS: macOS 15.1
    CPU: (12) arm64 Apple M2 Max
    Memory: 1.50 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 130.0.6723.71
    Safari: 18.1
  npmPackages:
    @rspress/plugin-client-redirects: ^1.28.2 => 1.32.0 

### Details

Type-based doc generation tools e.g. Typedoc, may render title with TS generic syntax `Component\<P, S\>` and will cause the inner markdown renderer of sidebar menu crush:

![Image](https://github.com/user-attachments/assets/36c43ff5-9516-45a1-9438-87dbdd41cba4)

### Reproduce link

N/A

### Reproduce Steps

Have markdown titled as:

```md
# Class: Component\<P, S\>
```

Huxpro:
Besides of fixing this particular issues, we should use more `ErrorBoundary` to prevent the page from being completely blank

Huxpro:
Emm another weird case

![Image](https://github.com/user-attachments/assets/113eb6fa-57a6-4f64-bcff-7adf3596f9e6)

Timeless0911:
> Type-based doc generation tools e.g. Typedoc, may render title with TS generic syntax Component\<P, S\> and will cause the inner markdown renderer of sidebar menu crush:

This is because in the parsing process of htmr when rending inline markdown text, `<P` has a special meaning in htmr, you can replace it with other variable names `<Z` maybe.

I prefer to use `` when rending code like text.

```mdx
# Class: `Component<d, S, SS>`
```

![Image](https://github.com/user-attachments/assets/b85430e7-0abd-447c-ba20-0dba9c357222)

Huxpro:
> > Type-based doc generation tools e.g. Typedoc, may render title with TS generic syntax Component\<P, S\> and will cause the inner markdown renderer of sidebar menu crush:
> 
> This is because in the parsing process of htmr when rending inline markdown text, `<P` has a special meaning in htmr, you can replace it with other variable names `<Z` maybe.
> 
> I prefer to use `` when rending code like text.
> 
> ```mdx
> # Class: `Component<P, S, SS>`
> ```
> 
[...]

PR #1557, opened 2024-11-06
fix: render inline markdown with generic syntax

Timeless0911:
## Summary

closes: #1536 

## Checklist


SoonIter (approved):

SoonIter:
😄 we should unify the sidebar, title, TOC, render logic

Timeless0911:
If we use `dangerouslySetInnerHTML`, it forces users to write HTML entities `&lt;` instead of `\<` which is more common in mdx writing. And the sidebar / markdown headings in cotent / TOC render logic should be unified.

PR #1563, opened 2024-11-07
fix: aside text should break word

Timeless0911:
### Before

image

### After

image

## Checklist

### DIFF
--- a/packages/theme-default/src/components/Aside/index.scss
+++ b/packages/theme-default/src/components/Aside/index.scss
@@ -20,4 +20,5 @@
   padding: 0px 12px;
   font-size: 0.8125rem;
   line-height: 1.25rem;
+  word-break: break-all;
 }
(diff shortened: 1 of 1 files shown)


fi3ework (approved):

packages/theme-default/src/components/Aside/index.scss
```diff
@@ -20,4 +20,5 @@
   padding: 0px 12px;
   font-size: 0.8125rem;
   line-height: 1.25rem;
+  word-break: break-all;
```
chenjiahan, line 23: I prefer using `overflow-wrap: break-word;`

chenjiahan: https://github.com/web-infra-dev/rspress/pull/1363

Timeless0911: got it

chenjiahan (approved):

PR #1597, opened 2024-11-18
chore: downgraden pnpm to 9.12.0 to avoid renovate break

Timeless0911:
## Summary

https://github.com/pnpm/pnpm/issues/7951#issuecomment-2438065169

Latest version of pnpm changed some behaviour, and the renovate generated PR will remove `pnpmfileChecksum` filed in lockfile which will cause no frozen lockfile error that breaks CI.

image

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "engines": {
     "node": ">=16.18.1"
   },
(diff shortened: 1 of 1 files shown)


Timeless0911:
seems to do not solve this issue.

PR #1602, opened 2024-11-18
fix: some remark packages should be devDependencies

chenjiahan:
## Summary

Some remark packages should be devDependencies:

image

## Checklist


Timeless0911 (approved):

pnpm-lock.yaml
```diff
@@ -2644,6 +2644,11 @@ packages:
     cpu: [x64]
     os: [win32]
 
+  '@rsbuild/core@1.1.1':
```
SoonIter, line 2647: weird lockfile, could you unify the rsbuild/core version?

Timeless0911 (approved):

PR #1617, opened 2024-11-21
build: supersede Modern.js Module with Rslib

fi3ework:
## Summary

Migrate part of the packages to built by Rslib.

Remove `threads: true`, it's removed in v2 and defaults to `'threads'` now.

## Checklist


packages/create-rspress/package.json
```diff
@@ -24,14 +24,15 @@
...
-    "dev": "rslib build --watch",
+    "dev": "rslib build -w",
     "start": "node ./dist/index.js"
   },
   "dependencies": {
     "create-rstack": "1.0.9"
   },
   "devDependencies": {
     "@rslib/core": "0.1.0",
+    "@microsoft/api-extractor": "^7.48.0",
```
Timeless0911, line 35: no need to install at this package?

fi3ework: I removed the rslib can api-e from root package.json

Timeless0911 (approved):

packages/modern-plugin-rspress/rslib.config.ts
```diff
@@ -0,0 +1,16 @@
...
+export default defineConfig({
+  lib: [
+    {
+      dts: {
+        bundle: true,
+      },
+      format: 'cjs',
+      syntax: 'es2020',
+      output: {
+        sourceMap: { js: 'source-map' },
```
chenjiahan, line 12: I think we do not need to ship source map for Rspress packages. This can reduce install size.

@Timeless0911 What is your opinion

Timeless0911: Agree, but why we add sourcemap before?

fi3ework: All sourcemap removed.

Timeless0911 (approved):

PR #1618, opened 2024-11-21
feat: doc-layout components merge classes from props

### LINKED ISSUE #1612: [Feature]: doc layout components should merge className

Currently, all of the internal components used by the `DocLayout` will completely override the `className` property. [For example, ul will always have "list-disc pl-5 my-4 leading-7"](https://github.com/web-infra-dev/rspress/blob/c6301987c4a07f78651f0e61b814ef4ddb9fee13/packages/theme-default/src/layout/DocLayout/docComponents/list.tsx#L8). I think it should respect any `className` it gets from props, and attempt to merge them.

The motivation for this is that I wrote a custom rehype plugin that adds custom classes to MDX elements, but these are stripped away by the components.

1. The simplest solution would be to concatenate the props.

```tsx
export const Ul = (props: ComponentProps<'ul'>) => {
  return <ul {...props} className={`list-disc pl-5 my-4 leading-7 ${props.className}`} />;
};
```

However, this may add an extra space if there were no classes provided. 

2. A nice solution would be to use the common `cn` utility for merging TailwindCSS classes.

```tsx
import { clsx, type ClassValue } from "clsx"
import { twMerge } from "tailwind-merge"

export function cn(...inputs: ClassValue[]) {
  return twMerge(clsx(inputs))
}

export const Ul = (props: ComponentProps<'ul'>) => {
  return <ul {...props} className={cn("list-disc pl-5 my-4 leading-7", props.className)} />;
};
```

This would require additional packages to be installed, but they are small and might be useful in other places.

3. Finally, it may also be feasible to implement an internal implementation of `cn` to prevent additional dependencies.

```tsx
export function cn(...inputs: unknown[]) {
  return inputs.filter(Boolean).join(' ')
}

export const Ul = (props: ComponentProps<'ul'>) => {
  return <ul {...props} className={cn("list-disc pl-5 my-4 leading-7", props.className)} />;
};
```

If any of these solutions seem acceptable, I can open a corresponding PR.

ap0nia:
## Summary

Refactored the default MDX components used by `DocLayout` to properly merge the `className` with the component's default `className`.

## Notes

All changes should be entirely non-breaking because the user needs to opt-in to adding classNames to the HTML elements. Additionally, very little was changed for each component.

The one exception would be the [`Pre` component](https://github.com/ap0nia/rspress/blob/e386467ffabad3ae67dc2268b7eeabcf27a3666e/packages/theme-default/src/layout/DocLayout/docComponents/pre.tsx#L42). In order to properly merge `className` intended for the parent and children, the wrapping div with `className="language-bash"` was lifted up outside `renderChildren`. And `renderChildren` was moved to module level. It looks functionally the same, but I have not confirmed it yet...

## Related Issue

Closes #1612 

## Checklist

I'm not sure which tests should be updated...
There is no documentation on how the default MDX components work, so this doesn't need to be documented either.

- ~~[ ] Documentation updated (or not required).~~

### DIFF
--- a/packages/theme-default/src/layout/DocLayout/docComponents/hr.tsx
+++ b/packages/theme-default/src/layout/DocLayout/docComponents/hr.tsx
@@ -1,10 +1,17 @@
 import type { ComponentProps } from 'react';
 
+import clsx from '../../../utils/tailwind';
+
 export const Hr = (props: ComponentProps<'hr'>) => {
+  const { className, ...restProps } = props;
+
   return (
     <hr
-      {...props}
-      className="my-12 border-t border-solid border-divider-light"
+      {...restProps}
+      className={clsx(
+        'my-12 border-t border-solid border-divider-light',
+        className,
+      )}
     />
   );
 };
--- a/packages/theme-default/src/layout/DocLayout/docComponents/link.tsx
+++ b/packages/theme-default/src/layout/DocLayout/docComponents/link.tsx
@@ -3,19 +3,21 @@ import styles from './index.module.scss';
 import { Link } from '@theme';
 import { usePathUtils } from '../../../logic';
 
+import clsx from '../../../utils/tailwind';
+
 export const A = (props: ComponentProps<'a'>) => {
   const { href = '', className = '' } = props;
   const { normalizeLinkHref } = usePathUtils();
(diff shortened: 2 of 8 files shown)


SoonIter:
thanks for your contribution

closed by too many conflicts from V2 main branch

Issue #1351, opened 2024-08-26
[Feature]: Support `jsdoc-to-markdown` of `@rspress/plugin-api-docgen`


SoonIter:
### What problem does this feature solve?

table of jsdoc like https://github.com/jsdoc2md/jsdoc-to-markdown

### What does the proposed API look like?

another parseTool like https://github.com/web-infra-dev/rspress/issues/1290

Issue #1290, opened 2024-07-31
[Feature]: Support `react-docgen` in `@rspress/plugin-api-docgen`


SoonIter:
### What problem does this feature solve?

https://rspress.dev/plugin/official-plugins/api-docgen#apiparsetool

1. Maintainability, [`react-docgen`](https://github.com/reactjs/react-docgen
) is the officital tooling to generate type doc of react component in Facebook, and the last-update of [`react-docgen-typescript`](https://github.com/styleguidist/react-docgen-typescript) is 2021

2. More features, analyze nested types and generate documents of nested types, and we can use the playground for better debugging of component-document

![img_v3_02da_171de5b0-1ed6-4f51-838e-927763df203g](https://github.com/user-attachments/assets/77599f40-02d1-4193-b998-a3aa4919ccb5)

### What does the proposed API look like?

https://rspress.dev/plugin/official-plugins/api-docgen#apiparsetool

1. use `apiParseTool` option `react-docgen` to switch from `react-docgen-typescript` to `react-docgen` 
2. We hope to switch the default value of `apiParseTool` to react-docgen in Rspress 2.0 and generate the component document for our new member of Rstack, Rslib XD

chenjiahan:
I'm curious if we can add a new plugin to support `react-docgen`, this can prevent `@rspress/plugin-api-docgen` from becoming too heavy.

SoonIter:
> I'm curious if we can add a new plugin to support `react-docgen`, this can prevent `@rspress/plugin-api-docgen` from becoming too heavy.

we can mark `documentation` and `react-docgen-typescript` as optional peerDependencies

or we can remove `react-docgen-typescript` directly

or both

chenjiahan:
Yeah, optional peerDependencies is better than built-in

SoonIter:
https://github.com/rspack-contrib/storybook-rsbuild/pull/223

PR #1632, opened 2024-11-30
feat(plugin-api-docgen): support jsdoc-to-markdown(#1351)

gaoachao:
## Related Issue
#1351

## Checklist


SoonIter:
thanks for your contribution, could you add a e2e test case for this feature?

you can debug it with `npx playwright test --ui`

SoonIter:
😭sorry, `jsdoc-to-markdown` seems that  it can only be used for parsing js files...
 it can only be used for eating.

gaoachao:
Yeap, `jsdoc-to-markdown` can only process js files, not ts. But as a developer option, it also has some value. If we indicate in the document that `entries` needs to be js file.

SoonIter:
> Yeap, `jsdoc-to-markdown` can only process js files, not ts. But as a developer option, it also has some value. If we indicate in the document that `entries` needs to be js file.

If few people use this feature, it will increase the maintenance cost and package size. 
Someone proposed to me to use `jsdoc-to-markdown`. He may just want to have a very simple tooling to extract comments of files, `documentation` extract not all the specific fields.

So sorry about the confusing issue, you could try out https://github.com/web-infra-dev/rspress/issues/1290

PR #1639, opened 2024-12-04
chore: use `resolve.alias` instead of `source.alias`

Timeless0911:
## Summary

chore: use `resolve.alias` instead of `source.alias`

## Checklist


chenjiahan (approved):
👍

PR #1642, opened 2024-12-05
fix(theme): Overview component cannot be used when using base

SoonIter:
## Summary
empty Overview when using base together
image

## Checklist


Timeless0911 (approved):
We should review and reorg the functions using in Node and runtime in the future.

Issue #1432, opened 2024-09-24
[Bug]: `head` config doesn't work on dev mode


notzheng:
### Version

```sh
System:
    OS: Linux 6.1 Debian GNU/Linux 11 (bullseye) 11 (bullseye)
    CPU: (2) x64 AMD EPYC
    Memory: 1.99 GB / 4.01 GB
    Container: Yes
    Shell: Unknown
  npmPackages:
    rspress: ^1.0.0 => 1.6.2
```

### Details

```
// rspress.config.ts
export default defineConfig({
  // ...other config
  head: [
    '<meta name="description" content="Site Description" />'
  ]
});
```
`head` config doesn't work on dev mode but work on build.

![Google Chrome  ScreentShot at 2024-09-24 15 27 48](https://github.com/user-attachments/assets/e95b88b8-9921-4ae9-860e-2b0c70429e5e)

### Reproduce link

https://codesandbox.io/p/devbox/relaxed-golick-dtkvl3

### Reproduce Steps

follow reproduce link.

Timeless0911:
We currently only adding head in `renderPages` process when in production build. 

You can directly configure [`builderConfig.html.tags`](https://rsbuild.dev/config/html/tags#htmltags), it will work in both dev and production mode.

PR #1660, opened 2024-12-13
docs: `head` config only takes effect in production mode

9aoy:
## Summary

update docs, `head` config only takes effect in production mode.

## Checklist


chenjiahan (commented):
Can we make it work in dev?

9aoy:
> Can we make it work in dev?

dev can support better.

Timeless0911:
See https://github.com/web-infra-dev/rspress/issues/1432

I think we can redesign `header` config in Rspress 2.0 to make it work well in both dev and production mode. Maybe we can directly compose and pass header-related config in config/fronmatter to `html.tags` config in Rsbuild.

Timeless0911 (approved):

PR #1662, opened 2024-12-15
perf: remove `node-fetch` polyfill

chenjiahan:
## Summary

Remove node-fetch polyfill to reduce 7.45MB install size:

image

- In Node.js >= 18, `fetch` is builtin and enabled by default.
- In Node.js 16, `fetch` can be enabled by the `--experimental-fetch` flag, I have updated the `bin` file to enable it.
- In Rspress 2.0 we can remove this flag from the bin file.

## Other Change

This PR also removed `mdast-util-mdxjs-esm` from two packages because they never use it:

Screenshot 2024-12-15 at 10 15 05

## TODO

I think the remote search should be redesigned in Rspress 2.0 as it is now a ByteDance SCM only feature.

See: https://github.com/web-infra-dev/rspress/blob/a373b32a6c5fc4db4c11201e2ea1e06b10dd86bc/packages/core/src/node/searchIndex.ts#L42

## Checklist

### DIFF
--- a/packages/cli/bin/rspress.js
+++ b/packages/cli/bin/rspress.js
@@ -1,4 +1,4 @@
-#!/usr/bin/env node
+#!/usr/bin/env node --experimental-fetch
 import nodeModule from 'node:module';
 
 // enable on-disk code caching of all modules loaded by Node.js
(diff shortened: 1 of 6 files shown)


Timeless0911 (approved):

SoonIter:
it seems that it is not a correct passing-shabang-option way

https://app.netlify.com/sites/rspack/deploys/676b3fe3df52430008077d47

image

PR #1670, opened 2024-12-16
fix: add code wrapping support for shiki highlighting

### LINKED ISSUE #1619: [Feature]: enable code-wrapping for shiki-highlighting

It was possible to enable wrapping with shiki-highlighting by [conditionally adding "white-space: pre" to the code element](https://github.com/ap0nia/eden-query/blob/9a5f0dfe2fbf12b2fa383a9d1318c02839a8152f/packages/rspress-plugin-twoslash/src/theme/code/index.jsx#L58).

However, this could cause breaking changes for people that don't want this style applied to their shiki codeblock. Alternatively, a `data` attribute could be attached to the element, and the user can apply css targeting it.

1. If the code block should be wrapped, apply a `data-rp-code-wrapped=true` attribute to the element.
2. The user can write custom CSS for that data-attribute.

```css
code[data-rs-code-wrapped] {
  white-space: pre-wrap;
}
```

Alternatively, always apply the white-space style for them, but allow it to be overridden. Or do both strategies; always apply the white-space style, but add a data-attribute that's easy to override the default styles from CSS.

JounQin:
## Related Issue

close #1619

## Checklist

### DIFF
--- a/packages/theme-default/src/layout/DocLayout/docComponents/code/index.tsx
+++ b/packages/theme-default/src/layout/DocLayout/docComponents/code/index.tsx
@@ -51,7 +51,17 @@ export function Code(props: CodeProps) {
         );
       case 'shiki':
       default:
-        return <code {...props}></code>;
+        return (
+          <pre
+            {...props}
+            style={{
+              whiteSpace: codeWrap ? 'pre-wrap' : 'pre',
+              overflowX: 'auto',
+            }}
+          >
+            <code>{props.children}</code>
+          </pre>
+        );
     }
   };
 
--- a/e2e/tests/plugin-shiki.test.ts
+++ b/e2e/tests/plugin-shiki.test.ts
@@ -1,6 +1,5 @@
 import path from 'node:path';
 import { expect, test } from '@playwright/test';
-import fixture from '../fixtures/plugin-rss/fixture.json';
 import {
   getPort,
   killProcess,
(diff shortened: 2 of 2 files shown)


packages/theme-default/src/layout/DocLayout/docComponents/code/index.tsx
```diff
@@ -51,7 +51,17 @@ export function Code(props: CodeProps) {
         );
       case 'shiki':
       default:
-        return <code {...props}></code>;
+        return (
+          <pre
```
JounQin, line 55: Align with `PrismSyntaxHighlighter` https://github.com/react-syntax-highlighter/react-syntax-highlighter/blob/842b45ba9328cc232cf6d085e2e6db97291b797d/src/highlight.js#L432

packages/theme-default/src/layout/DocLayout/docComponents/code/index.tsx
```diff
@@ -51,7 +51,17 @@ export function Code(props: CodeProps) {
...
+            {...props}
+            style={{
+              whiteSpace: codeWrap ? 'pre-wrap' : 'pre',
+              overflowX: 'auto',
```
JounQin, line 59: Without this style, too long lines will overflow the whole container, and the code button group will not work as expected:

image

Timeless0911 (approved):

Issue #1696, opened 2024-12-23
[Bug]: navbar with hash link doesn't change URL when click


coratgerl:
### Version

```sh
System:
    OS: macOS 15.1.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 2.54 GB / 32.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Brave Browser: 121.1.62.156
    Chrome: 131.0.6778.205
    Safari: 18.1.1
  npmPackages:
    rspress: workspace:* => 1.38.0
```

### Details

When you configure nav to redirect to a section with #id rspress correctly scroll to the section but the url doesn't change. It can cause issue when we want to track where the user is on the page with tools like plausible or others.

### Reproduce Steps

Here click on **PageA** redirect to /index#pageA, then click on **PageB** scroll to pageB section but keep the url /index#pageB

```typescript
export default defineConfig({
  root: path.join(__dirname, 'doc'),
  route: {
    cleanUrls: true,
  },
  themeConfig: {
    nav: [
      {
        text: 'PageA',
        link: '#pageA',
        position: 'right',
      },
      {
        text: 'PageB',
        link: '#pageB',
        position: 'right',
      },
    ],
  },
});

```

PR #1697, opened 2024-12-23
fix: navbar link should change url when click on item with hash

coratgerl:
## Summary

Click on an item that link to an hashtag section should change the url when click. This is useful when we want for example to track where the user click with tools like plausible or unami.

## Related Issue

Fixes: https://github.com/web-infra-dev/rspress/issues/1696

## Checklist


Timeless0911 (approved):
LGTM

PR #1699, opened 2024-12-24
chore(CI): modify release steps

Timeless0911:
## Summary

modify release steps

## Checklist

### DIFF
--- a/scripts/generateReleasePr.mjs
+++ b/scripts/generateReleasePr.mjs
@@ -1,0 +1,106 @@
+#!/usr/bin/env zx
+
+import fs from 'node:fs/promises';
+import path from 'node:path';
+import { parseArgs } from 'node:util';
+import { $, chalk } from 'zx';
+
+$.verbose = false;
+
+async function getCurrentVersion() {
+  const packageJsonPath = path.join(
+    process.cwd(),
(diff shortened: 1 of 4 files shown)


scripts/generateReleasePr.mjs
```diff
@@ -0,0 +1,106 @@
...
+  const packageJsonPath = path.join(
+    process.cwd(),
+    'packages/core/package.json',
+  );
+  const packageJson = JSON.parse(await fs.readFile(packageJsonPath, 'utf-8'));
+  return packageJson.version;
+}
+
+async function getNextVersion(currentVersion, type) {
+  const [major, minor, patch] = currentVersion.split('.').map(Number);
```
chenjiahan, line 20: alpha / beta / rc not supported?

Timeless0911: If we use changeset to generate `alpha / beta / rc` version, we should use prerelease mode and `patch/minor/major` changeset is also needed. This script can automatically generate a changeset file and consume it. This is a bit complicated now.

chenjiahan: Get, we can improve the workflow for prereleases in the future.

chenjiahan (approved):

PR #1770, opened 2025-01-24
chore: migrate packages building with Rslib

fi3ework:
## Summary

Migrate @rspress/plugin-playground and @rspress/plugin-preview to build with Rslib.

### TODO: 

**update:** 

After discussion, we decide to not add a new option as `output.externals` takes effect for dts output. We'll add the method into Q&A chapter of Rslib.

--- 

The build has failed because we are missing the following items:

`react` is declared in `devDependencies` so it tends to be bundled while we do not want. We need a new dts configuration like `dts.externals` (or other name) to customize excluding packages from being bundled, and covers the config https://modernjs.dev/module-tools/en/api/config/build-config.html#dtsrespectexternal of Modern.js Module.

The config type might be like below:

```ts
type DtsExternals = string[] | (string[]) => string[]
```

@Timeless0911 

## Checklist


SoonIter (approved):
LGTM

PR #1790, opened 2025-01-28
chore(infra): implement click to reference by tsconfig.reference

SoonIter:
## Summary

this is a trick for implementing "click to reference" of ts-language-server

tsc will build the `.d.ts.map` in memory

so we need to set `rootDir: "src"` `"outDir: "dist""` `composite: true` and tsconfig.reference just like Rsbuild repo

img_v3_02iv_0cb9e632-dd84-465c-bf86-39fe0f25b46g

BTW, fix some type error when debugging references

## Checklist

### DIFF
--- a/packages/cli/src/update.ts
+++ b/packages/cli/src/update.ts
@@ -1,7 +1,7 @@
 import { spawn } from 'node:child_process';
 import fs from 'node:fs/promises';
 import path from 'node:path';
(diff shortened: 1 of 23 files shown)


Timeless0911 (approved):

Timeless0911:
Check whether the DTS output of each package is as expected.

Issue #1543, opened 2024-11-02
[Feature]: Code blocks are searchable.


jheng-jie:
### What problem does this feature solve?

Some feature descriptions may be included within code blocks, and sometimes .md files are generated automatically, identifying content from comments during generation. Therefore, the search functionality really needs to be able to search within code blocks.

### What does the proposed API look like?

Make it configurable in the settings file.

Timeless0911:
We will complete this feature in the development of v2 version, and PR is welcome if you are interested.

PR #1797, opened 2025-02-02
feat: support for searching code blocks

chenjiahan:
## Summary

This PR adds a new `search.codeBlocks` option.

If enabled, the search index will include code block content, which allows users to search code blocks.

```ts title="rspress.config.ts"
import { defineConfig } from 'rspress/config';

export default defineConfig({
  search: {
    codeBlocks: true,
  },
});
```

## Default value

This default value is `false` to align with other search engines like Docsearch, see: https://docsearch.algolia.com/docs/docsearch-program/#can-you-index-code-samples.

If it works well, we may consider enabling it by default in the future.

## Related Issue

resolve https://github.com/web-infra-dev/rspress/issues/1543

## Checklist


packages/core/src/node/runtimeModule/siteData/extractPageData.ts
```diff
@@ -108,7 +109,7 @@ export async function extractPageData(
             {
               // Skip code blocks
               selector: 'pre > code',
-              format: 'skip',
+              format: searchCodeBlocks ? 'inlineTag' : 'skip',
```
SoonIter, line 112: why uses 'inlineTag' here

Is `<code />` under expectation?

image


info: https://github.com/html-to-text/node-html-to-text/blob/master/packages/html-to-text/README.md

chenjiahan: All in built-it formats of `html-to-text` will preserve the `<code>` prefix.

I guess this is not bad, since it shows that the search results are from code blocks.

If you have time, you can try writing a custom formatter to remove `<code>` 😄

SoonIter: > All the built-it formats of html-to-text will preserve the \<code\> prefix.

default format `format: 'block'` of `html-to-text` would remove the html tag (to-text). So I asked whether the preservation of \<code \/\> is deliberate.

> I guess this is not bad, since it shows that the search results are from code blocks.

Yes, this is useful to users, but some results contain it while others do not, which is strange. We can do this but in another way.


before

image


after

image

SoonIter:
```suggestion
              format: searchCodeBlocks ? 'block' : 'skip',
```

chenjiahan: LGTM 👍

SoonIter (approved):

PR #1832, opened 2025-02-10
chore(infra): svgr config modification

SoonIter:
## Summary

`query: /.*/` works, but I do not recommend this usage which is not written in doc

## Related Issue

https://github.com/web-infra-dev/rslib/pull/741

https://rsbuild.dev/zh/plugins/list/plugin-svgr#%E5%85%B7%E5%90%8D%E5%AF%BC%E5%85%A5

## Checklist


chenjiahan (approved):
👍

PR #1844, opened 2025-02-13
fix(core): search json path is incorrect when config.outDir is absolute

liuwenzhuang:
## Summary

Serve search json will be failed when `config.outDir` is a absolute path, use same path join strategy from `writeSearchIndex` function instead.

## Checklist


SoonIter (approved):

SoonIter:
Thanks❤, this indeed appears to be a corner case
Feel free to further test with the `config.outputDir: absolutePath` setting, submit PRs, and add more test cases to ensure stability of the functionality.

PR #1848, opened 2025-02-13
fix(core): supersede require with import statement in ESM output

fi3ework:
## Summary

Remove possible `require` statement reserved in ESM output, and clean Rslib config by the way.

## Checklist

### DIFF
--- a/packages/core/src/node/runtimeModule/siteData/highlightLanguages.ts
+++ b/packages/core/src/node/runtimeModule/siteData/highlightLanguages.ts
@@ -1,12 +1,14 @@
+import { createRequire } from 'node:module';
 import { DEFAULT_HIGHLIGHT_LANGUAGES } from '@rspress/shared';
 import picocolors from 'picocolors';
 
+const require = createRequire(import.meta.url);
 let supportedLanguages: Set<string>;
 
 export function handleHighlightLanguages(
   highlightLanguages: Set<string>,
   defaultLanguages: (string | [string, string])[],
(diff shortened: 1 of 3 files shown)


packages/core/src/node/runtimeModule/siteData/highlightLanguages.ts
```diff
@@ -1,12 +1,14 @@
...
 import { DEFAULT_HIGHLIGHT_LANGUAGES } from '@rspress/shared';
 import picocolors from 'picocolors';
 
+const require = createRequire(import.meta.url);
 let supportedLanguages: Set<string>;
 
 export function handleHighlightLanguages(
   highlightLanguages: Set<string>,
   defaultLanguages: (string | [string, string])[],
-): Record<string, string[]> {
```
Timeless0911, line 9: Why remove this return type?

packages/core/src/node/runtimeModule/siteData/highlightLanguages.ts
```diff
@@ -1,12 +1,14 @@
...
+) {
```
fi3ework, line 11:
```suggestion
): Record<string, string[]> {
```

packages/core/src/node/runtimeModule/siteData/highlightLanguages.ts
fi3ework: Restored.

Timeless0911 (approved):

PR #1850, opened 2025-02-14
refactor: should emit the detailed error stack

SoonIter:
## Summary

should emit the full error with error.stack

## Checklist

### DIFF
--- a/packages/core/src/node/utils/checkLanguageParity.ts
+++ b/packages/core/src/node/utils/checkLanguageParity.ts
@@ -73,6 +73,7 @@ async function collectModuleFiles(
       fileLangMap[baseName].add(lang);
     }
   } catch (e) {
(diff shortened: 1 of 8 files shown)


packages/core/src/node/utils/checkLanguageParity.ts
```diff
@@ -143,7 +144,7 @@ export async function checkLanguageParity(config: UserConfig) {
     }
     logger.success('Language parity checked successfully.');
   } catch (err) {
-    logger.error(`Error during language parity check: ${err.message}`);
+    logger.error('Error during language parity check: ', err);
```
Timeless0911, line 147:
```suggestion
    logger.error('Error during language parity check.);
    logger.error(err);
```

Timeless0911 (approved):

PR #1856, opened 2025-02-17
fix(ui): when click the url, calculating hash in real time

SoonIter:
## Summary

fix(ui): when click the url, calculating hash in real time

this may fix the jump and scroll issue

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #1858, opened 2025-02-17
fix(ui): for TOC highlight, revert `setTimeout` set unbinding changes

SoonIter:
## Summary

revert some changes of https://github.com/web-infra-dev/rspress/pull/1856

after further investigation, 

1. why scroll to the middle content when navigating to other page? 

` history.scrollRestoration = 'auto'`

https://github.com/web-infra-dev/rspress/pull/1181

2. why revert `setTimeout` related code?

there is a TOC highlight bug

https://github.com/user-attachments/assets/32d61e32-69e9-4f52-8f48-b9c0f84f3013

3. interesting things

https://github.com/web-infra-dev/rspress/blob/48df322887efa2975ac4feb0d80ae48ea4522acf/packages/theme-default/src/logic/sideEffects.ts#L182 

these codes are useless 😄

we currently all the scroll behavior is in 

https://github.com/web-infra-dev/rspress/blob/48df322887efa2975ac4feb0d80ae48ea4522acf/packages/theme-default/src/components/Aside/index.tsx#L69-74

 should govern them

## Checklist

### DIFF
--- a/packages/theme-default/src/components/Aside/index.tsx
+++ b/packages/theme-default/src/components/Aside/index.tsx
@@ -14,8 +14,7 @@ import { useLocation } from '@rspress/runtime';
 const TocItem = ({
   header,
   baseHeaderLevel,
-  hiddenNav,
-}: { header: Header; baseHeaderLevel: number; hiddenNav: boolean }) => {
+}: { header: Header; baseHeaderLevel: number }) => {
   return (
     <li>
       <a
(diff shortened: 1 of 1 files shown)


packages/theme-default/src/components/Aside/index.tsx
```diff
@@ -29,10 +28,6 @@ const TocItem = ({
         onClick={e => {
           e.preventDefault();
           window.location.hash = header.id;
-          const target = document.getElementById(header.id);
```
SoonIter, line 32: this scroll code is needless because we do the scroll in `useEffect`

Timeless0911 (approved):

SoonIter:
revert first, make main safe

refactor later 😄

too fragile to modify these codes

PR #1863, opened 2025-02-18
refactor(theme): `Sidebar` isActive algorithm should be same with `Route` match

SoonIter:
## Summary

We currently have two matching algorithms for sidebar

```ts
{
	'/': [],
	'/guide': [
		{
			text: 'Getting Started',
			link: '/guide/getting-started',
		},
	],
},
```
#### 1. `matchPath(sidebarKey, pathname)` in `getSidebarGroupData(pathname: string)`

decide which **sidebarGroup** would show, and it will calculate with these keys `'/'`, `'/guide'` one by one

https://github.com/web-infra-dev/rspress/blob/fb81b996c5630b826ee6f5df7e1a6920d9d48e52/packages/theme-default/src/logic/getSidebarDataGroup.ts#L22-L42

#### 2. `match(sidebarGroup, pathname)`

decide which **sidebarItem** is active,

to match `item.link` recursively

```ts
{
	text: 'Getting Started',
	link: '/guide/getting-started',
},
```

## some code changes

1. unify all `pathname -> routePath` logic to `pathnameToRouteService`

it is a core function which bridge `Route` and `Sidebar`

2. use `pathnameToRouteService` to decide which sidebar is active

In rspress, we calculating it repeatedly

This  `pathnameToRouteService`  should be the above algorithm 2,  it comes froms `react-router-dom`

Of course, Sidebar `isActive` should use the same logic with Route

3. be compatible to `api-extractor` generated path

for rspeedy

## Checklist

### DIFF
--- a/packages/plugin-preview/static/global-components/common/mobile-operation.tsx
+++ b/packages/plugin-preview/static/global-components/common/mobile-operation.tsx
@@ -57,11 +57,6 @@ export default (props: {
 
   const onClickOutside = useCallback(
     (e: MouseEvent) => {
-      console.log(
-        !contains(triggerRef.current, e.target),
-        triggerRef.current,
-        e.target,
-      );
       if (!contains(triggerRef.current, e.target)) {
         setShowQRCode(false);
       }
--- a/packages/core/src/runtime/App.tsx
+++ b/packages/core/src/runtime/App.tsx
@@ -41,7 +41,10 @@ export async function initPageData(routePath: string): Promise<PageData> {
         p
           .replace(/\/$/, '')
           .toLowerCase();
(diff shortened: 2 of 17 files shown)


packages/plugin-preview/static/global-components/common/mobile-operation.tsx
```diff
@@ -57,11 +57,6 @@ export default (props: {
 
   const onClickOutside = useCallback(
     (e: MouseEvent) => {
-      console.log(
```
SoonIter, line 60: 😄 funny

Timeless0911 (approved):

Issue #1864, opened 2025-02-18
[Bug]: Code search feature can't search component format code.


KyrieLii:
### Version

```sh
latest
```

### Details

<img width="1274" alt="Image" src="https://github.com/user-attachments/assets/eaeaafbd-71e4-41e1-b28c-9d08d040e843" />

### Reproduce link

https://github.com/web-infra-dev/rspress/blob/main/e2e/fixtures/search-code-blocks/doc/index.mdx

### Reproduce Steps

Add to index.mdx:
```
<React.Suspense>
      <Logo />
    </React.Suspense>
```

then search `React`.

PR #1867, opened 2025-02-19
fix(search): should search JSX element code block

SoonIter:
## Summary

fix(search): should search JSX element code block

## Related Issue

close https://github.com/web-infra-dev/rspress/issues/1864

## Checklist

### DIFF
--- a/packages/core/src/node/runtimeModule/siteData/extractPageData.ts
+++ b/packages/core/src/node/runtimeModule/siteData/extractPageData.ts
@@ -97,7 +97,7 @@ export async function extractPageData(
         }
 
         /**
-         * escape jsx element in code block, or it cannot be searched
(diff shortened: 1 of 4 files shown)


packages/core/src/node/runtimeModule/siteData/extractPageData.ts
```diff
@@ -96,7 +96,26 @@ export async function extractPageData(
           return null;
         }
 
-        content = htmlToText(String(html), {
+        /**
+         * escape jsx element in code block, or it cannot be searched
```
chenjiahan, line 100:
```suggestion
         * Escape JSX elements in code block to allow them to be searched
```

Timeless0911 (dismissed):

PR #1878, opened 2025-02-25
perf(size): 🔥remove package.json useless field and bundle named package.json

SoonIter:
## Summary

1. update `"files"` list, remove src and useless files

2. bundle package.json with namedImport

3. remove `"jsnext: source"` field

## Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -15,8 +15,7 @@
     "*.sass",
     "*.scss",
     "**/virtual-global-styles.js",
-    "virtual-global-styles",
-    "./src/theme-default/styles/index.ts"
+    "virtual-global-styles"
   ],
   "type": "module",
   "exports": {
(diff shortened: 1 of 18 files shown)


packages/core/src/node/build.ts
```diff
@@ -12,6 +12,7 @@ import {
 import { logger } from '@rspress/shared/logger';
 import picocolors from 'picocolors';
 import { HelmetData } from 'react-helmet-async';
+import { version } from '../../package.json';
```
SoonIter, line 15: 1. we have already imported package.json once, no need `await import`

2. 😄 one js object, no need N * promises object to do `await import`

```js
const pkgJson = {version: "1.42.0"}
```

packages/core/package.json
```diff
@@ -29,7 +28,6 @@
     "./theme": {
       "default": "./theme.ts"
     },
-    "./theme.css": "./theme.css",
```
SoonIter, line 32: should removed in https://github.com/web-infra-dev/rspress/pull/713

Timeless0911 (approved):

PR #1879, opened 2025-02-25
refactor(tsconfig): migrate tsconfig to @rspress/config/tsconfig

SoonIter:
## Summary

maybe remove `"noUnusedLocals": false` in anothor pr

1. migrate tsconfig to @rspress/config/tsconfig from @modern-js/tsconfig

2. delete vitest.config.ts, we have already used the vitest.workspace.ts

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #1882, opened 2025-02-26
refactor(types): turn on `"strict": true` for @rspress/core

SoonIter:
## Summary

1. https://github.com/web-infra-dev/rspress/pull/1879
2. https://github.com/web-infra-dev/rspress/pull/1880

## Checklist


packages/core/src/runtime/App.tsx
```diff
@@ -31,12 +32,13 @@ type PageMeta = {
 };
 
 export async function initPageData(routePath: string): Promise<PageData> {
-  const matchedRoute = pathnameToRouteService(routePath)!;
+  const matchedRoute = pathnameToRouteService(routePath);
+  console.log(siteData, 22222222);
```
Timeless0911, line 36: 1

packages/theme-default/src/logic/useLocaleSiteData.ts
```diff
@@ -6,6 +6,9 @@ export function useLocaleSiteData(): NormalizedLocales {
   const {
     page: { lang },
   } = pageData;
+  if (lang === undefined) {
+    console.log(pageData, 111111, lang);
```
Timeless0911, line 10: 1

Timeless0911 (approved):

packages/core/src/node/mdx/loader.ts
```diff
@@ -211,9 +222,11 @@ MDXContent.__RSPRESS_PAGE_META["${encodeURIComponent(
     )}"] = ${JSON.stringify(pageMeta)};
 `;
     callback(null, result);
-  } catch (e) {
-    logger.error(`MDX compile error: ${e.message} in ${filepath}`);
-    logger.debug(e);
-    callback({ message: e.message, name: `${filepath} compile error` });
+  } catch (e: unknown) {
```
markH5, line 225: In js, accidentally throwing something other than an `error` is a common bad practice.

Even use tools like eslint in project to avoid it, it's still hard to avoid upstream dependencies that throw things other than error.

If there are types other than error, is it more robust to keep throwing errors, or is it more robust to have any records at all?

SoonIter: 😂 I do not know, 
IMO, "be robust" is not what we should pursue, panic and error message is also a good choice to help us debug

SoonIter: you can have a further investigation, to search how other libraries do it

PR #1893, opened 2025-03-02
perf(search): remove group calculation

chenjiahan:
## Summary

The `getSidebarDataGroup` can be very slow when there are thousands of routes.

This PR removed the group calculation and use `item.title` to group the search results.

In the future, the sorting of search results should be better implemented (for example, according to some weight indicators we agree on, such as title > content > code).

## Related PR

Performance regression since https://github.com/web-infra-dev/rspress/pull/1863

## Checklist

### DIFF
--- a/packages/theme-default/src/components/Search/SearchPanel.tsx
+++ b/packages/theme-default/src/components/Search/SearchPanel.tsx
@@ -11,7 +11,6 @@ import { debounce } from 'lodash-es';
 import { useCallback, useEffect, useRef, useState } from 'react';
 import { createPortal } from 'react-dom';
 import * as userSearchHooks from 'virtual-search-hooks';
-import { getSidebarDataGroup } from '../../logic/getSidebarDataGroup';
 import { useLocaleSiteData } from '../../logic/useLocaleSiteData';
 import { SvgWrapper } from '../SvgWrapper';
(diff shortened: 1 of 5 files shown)


packages/theme-default/src/components/Search/SearchPanel.tsx
```diff
@@ -353,7 +346,7 @@ export function SearchPanel({ focused, setFocused }: SearchPanelProps) {
   ): Record<string, DefaultMatchResultItem[]> => {
     return suggestions.reduce(
       (groups, item) => {
-        const group = item.group;
+        const group = item.title;
```
SoonIter, line 349: so the suggestions is not grouped after this change?

chenjiahan: Now grouped by `item.title`

chenjiahan: In most cases the search results are the same as before

SoonIter: ### before

grouped by navBar item

image

### after

grouped by page (title)

image

SoonIter (approved):
### before

grouped by navBar item

image

### after

grouped by page (title)

image

PR #1899, opened 2025-03-02
perf(search): improve initialization and prefetching logic

chenjiahan:
## Summary

For large sites, initializing the search while the page is idle may cause the page to render slowly.

To address this issue, this PR splits search init into two steps: prefetch and init:

- prefetch the search index when the page is idle
- init the search when the search bar is focused.

## Checklist


SoonIter (approved):
not the real perf😂, just UX perf

PR #1921, opened 2025-03-07
chore: use pnpm instead of npx

JounQin:
## Summary

Otherwise, it'll result the following warning for `npm@11` installed.

```log
npm warn Unknown project config "strict-peer-dependencies". This will stop working in the next major version of npm.
npm warn Unknown project config "link-workspace-packages". This will stop working in the next major version of npm.
```

## Related Issue

N/A

## Checklist


SoonIter (approved):
thanks ღ( ´･ᴗ･` )

PR #1914, opened 2025-03-06
fix: display outline correctly on >=960px <=1280px screen width

JounQin:
## Related Issue

close #1540

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

---

image

image

PR #1928, opened 2025-03-10
fix: rework sidebar menu responsive issues

### LINKED ISSUE #1540: [Bug]: `outline` does not display when window width `>= 960 && < 1280`
### Version

  System:
    OS: macOS 15.0.1
    CPU: (10) arm64 Apple M1 Max
    Memory: 3.53 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 130.0.6723.92
    Safari: 18.0.1
  npmPackages:
    @rspress/core: ^1.35.2 => 1.35.2 

### Details

Image

I recommend to render TOC in document in this case or even for mobile mode.

Image

### Reproduce link

https://rspress.dev/guide/start/introduction

### Reproduce Steps

N/A

### LINKED ISSUE #1924: [Feature]: Enhance code block responsive on  mobile device

The mobile compatibility isn't good if the code block wider than mobile screen width.
Image

The code block should support horizontal scroll to fix the mobile compatibility.

JounQin:
## Related Issue

close #1540

related #1914

## Checklist

---

I aslo added `--rp-content-padding-x` and `--rp-content-padding-y` css variables so that the aside can be responsive their changes.

### DIFF
--- a/packages/theme-default/src/logic/utils.ts
+++ b/packages/theme-default/src/logic/utils.ts
@@ -1,7 +1,7 @@
 import htmr from 'htmr';
 
 export function isMobileDevice() {
-  return window.innerWidth <= 1024;
+  return window.innerWidth < 1280;
 }
 
 export function renderHtmlOrText(str?: string | number | null) {
--- a/e2e/tests/hide-nav-bar.test.ts
+++ b/e2e/tests/hide-nav-bar.test.ts
@@ -6,7 +6,6 @@ const fixtureDir = path.resolve(__dirname, '../fixtures');
 async function isNavBarVisible(page: Page): Promise<boolean> {
   const nav = await page.$('.rspress-nav');
   const className: string = await nav?.evaluate(el => el.className);
-
   return !className.includes('hidden');
 }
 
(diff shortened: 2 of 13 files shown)


SoonIter:
:D thanks for your fix, could you take a look at the 3 problem in https://github.com/web-infra-dev/rspress/pull/1929?

SoonIter:
Because this issue has some impact, I have decided to revert it first in https://github.com/web-infra-dev/rspress/pull/1930 and add it back later.

packages/theme-default/src/logic/utils.ts
```diff
@@ -1,7 +1,7 @@
 import htmr from 'htmr';
 
 export function isMobileDevice() {
```
SoonIter, line 3: Is it possible for us to remove `isMobileDevice`? In my opinion, all device judgments can be made through CSS media query.

SoonIter: it might introduce the the 3rd problem in https://github.com/web-infra-dev/rspress/pull/1929

JounQin: It's used with checking `siteData.themeConfig.hideNavbar` together:

```ts
const computeNavPosition = () => {
    // On doc page we have the menu bar that is already sticky
    if (!isMobile) {
      return 'sticky';
    }
    if (siteData.themeConfig.hideNavbar === 'never' && page.pageType !== 'doc')
      return 'sticky';

    return 'relative';
  };
```

SoonIter: thanks, codes looks nicer than before

SoonIter (approved):

PR #1947, opened 2025-03-12
Release v2.0.0-alpha.2

SoonIter:
## Checklist


.changeset/1741687770983-alpha-release.md
```diff
@@ -0,0 +1,5 @@
+---
```
Timeless0911, line 1: should delete this in release script?

SoonIter: we can delete all when publishing v2.0.0

PR #1966, opened 2025-03-17
fix(style): unset the navItem bg color

SoonIter:
### Before

image

### After

image

remove compact

image

## Checklist


chenjiahan (commented):
I think we can make the nav items more compact by default after this change:

image

chenjiahan (approved):

PR #1969, opened 2025-03-17
docs: add `editLink` and `outline` config

9aoy:
## Summary

Add configuration for `editLink` and `outline` to the doc based on the type definition.

BTW, `docFooter` is not documented and does not seem to be used.

https://github.com/web-infra-dev/rspress/blob/a9cdd91364ba44a80b149f09a862b9ac34e70d75/packages/shared/src/types/defaultTheme.ts

## Checklist


SoonIter (approved):
thanks🫀

PR #1975, opened 2025-03-18
chore(theme/a11y): a11y for lighthouse score

markH5:
## Summary

before
> before

after
> after

## Checklist


packages/theme-default/src/components/Nav/NavMenuGroup.tsx
```diff
@@ -123,6 +123,7 @@ export function NavMenuGroup(item: NavMenuGroupItem) {
   return (
     <div className="relative flex-center h-14" onMouseLeave={handleMouseLeave}>
       <button
+        aria-label="change-lang"
```
SoonIter, line 126: this label content is not correct

PR #1939, opened 2025-03-11
fix(mdxjs): not need to add base on normalizing links with base enabled

JounQin:
## Related Issue

`pnpm dev:website` emit error:

```log
error   Internal link to /test/en/api/ is dead, check it in en/api/commands.mdx
```

image

See also build error https://app.netlify.com/sites/aquamarine-blini-95325f/deploys/67d01cec5b8a620008961b67

I tried to add logs for

`internalLinks`:

https://github.com/web-infra-dev/rspress/blob/69a140ec814ded5c79e0ca38fa6f77432cc4265a/packages/core/src/node/mdx/remarkPlugins/checkDeadLink.ts#L77

```log
Set(1) { '/test/en/api/' } <ROOT>/en/api/commands.mdx <ROOT>
error   Internal link to /test/en/api/ is dead, check it in en/api/commands.mdx
```

`links`:

https://github.com/web-infra-dev/rspress/blob/69a140ec814ded5c79e0ca38fa6f77432cc4265a/packages/core/src/node/mdx/loader.ts#L207

```log
[
  '#commands',
  '/en/api/index',
  '#rspress-dev',
  '#rspress-build',
  '#rspress-preview',
  '#rspress-update'
] <ROOT>/en/api/commands.mdx <ROOT>
```

What means `internalLinks` has `base` incorrectly if compared with `links`, that happens at https://github.com/web-infra-dev/rspress/blob/69a140ec814ded5c79e0ca38fa6f77432cc4265a/packages/core/src/node/mdx/remarkPlugins/normalizeLink.ts#L53

## Checklist

- [x] Tests updated (or not required).
- [x] Documentation updated (or not required).

Issue #1916, opened 2025-03-06
[Feature]: Refactor normalizeHref for Optimal SEO URL Handling


coratgerl:
### What problem does this feature solve?

Currently, the normalizeHref function in the rspress repository transforms URLs in a way that may not be optimal for SEO. Specifically, it converts / to /index for clean URLs and / to /index.html for non-clean URLs. This behavior can lead to suboptimal SEO performance because canonical URLs (shorter and cleaner) are generally better for search engine rankings.

**Current Behavior:**

Clean URLs: / -> /index
Non-clean URLs: / -> /index.html

**Proposed Changes:**

Refactor the normalizeHref function to always use the root URL / instead of appending index or index.html.
Consider removing the cleanUrls option as it introduces non-optimal behavior from an SEO perspective.

**Benefits:**
- Improved SEO by using canonical URLs.
- Simplified URL structure, making it easier for users and search engines to understand.
- Removal of redundant cleanUrls option, reducing potential confusion and misuse.

If you are ok I can create a PR for this changes

Timeless0911:
IMO, I think we also need to sort out and unify the concatenation of `version` / `lang` / `/index` / `cleanUrl` in the url. This is a more complicated matter and may require a certain amount of refactoring.

PR #1988, opened 2025-03-21
fix(mdxjs): add `version` and `lang` part when normalizing links

JounQin:
## Related Issue

Another critical issue similar to #1939

Without correct `version` and `lang` prefixes, `/no-en.mdx` link resolves `/no-en` in `/zh` routes, it renders fine, but `checkDeadLinks` doesn't know `/no-en` but `/zh/no-en`, so adding the prefixes are required.

Reproduction: https://app.netlify.com/sites/rspress-v2/deploys/67dcf3049a82a60008225ec8

image

cc @SoonIter @Timeless0911 

## Checklist

### DIFF
--- a/packages/core/src/node/mdx/remarkPlugins/normalizeLink.ts
+++ b/packages/core/src/node/mdx/remarkPlugins/normalizeLink.ts
@@ -54,7 +54,7 @@ export const remarkPluginNormalizeLink: Plugin<
       if (
         (routeService?.extensions ?? DEFAULT_PAGE_EXTENSIONS).includes(extname)
       ) {
-        url = url.replace(extname, '');
+        url = url.replace(new RegExp(`\\${extname}$`), '');
       }
 
       const relativePath = path.relative(root, file.path);
--- a/packages/core/src/node/mdx/remarkPlugins/normalizeLink.ts
+++ b/packages/core/src/node/mdx/remarkPlugins/normalizeLink.ts
@@ -62,18 +62,19 @@ export const remarkPluginNormalizeLink: Plugin<
       if (url.startsWith('.')) {
         url = path.posix.join(slash(path.dirname(relativePath)), url);
       } else if (routeService) {
-        url = addLeadingSlash(url);
-
         const [pathVersion, pathLang] = routeService.getRoutePathParts(
           slash(relativePath),
         );
-        const [urlVersion, urlLang] = routeService.getRoutePathParts(url);
+        const [urlVersion, urlLang, urlPath] =
+          routeService.getRoutePathParts(url);
 
-        if (!urlLang && pathLang) {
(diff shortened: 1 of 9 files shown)


packages/core/src/node/route/RouteService.ts
```diff
@@ -234,6 +238,16 @@ ${routeMeta
 `;
   }
 
+  getRoutePathParts(routePath: string) {
```
SoonIter, line 241: could you add some tests for getRoutePathParts just like `normalizeRoutePath`?

Timeless0911:
https://github.com/web-infra-dev/rspress/issues/1916#issuecomment-2705487505

Is the link normalization completely refactored and unified now?😂

SoonIter:
you can have a try to remove the `usePathUtils` here, it is for adding the `version` and `lang` before but in runtime

https://github.com/web-infra-dev/rspress/blob/12e42e9d5b1d658385db7bb605dbf4e698589a83/packages/theme-default/src/layout/DocLayout/docComponents/a.tsx#L15

packages/core/src/node/route/RouteService.ts
JounQin: Doesn't testing `normalizeRoutePath` alreay tested `getRoutePathParts`? But as you asked, I'll add that then.

JounQin:
> you can have a try to remove the `usePathUtils` here, it is for adding the `version` and `lang` before but in runtime
> 
> https://github.com/web-infra-dev/rspress/blob/12e42e9d5b1d658385db7bb605dbf4e698589a83/packages/theme-default/src/layout/DocLayout/docComponents/a.tsx#L15

@SoonIter This can not be dropped right now, because this PR fix the mdxjs side, for [mdxrs](https://github.com/web-infra-dev/mdx-rs/issues/74), I'll do in another PR and then remove `usePathUtils` eventually.

packages/core/src/node/mdx/remarkPlugins/normalizeLink.ts
```diff
@@ -36,13 +51,32 @@ export const remarkPluginNormalizeLink: Plugin<
 
       const extname = path.extname(url);
 
-      if (extname === '.md' || extname === '.mdx') {
-        url = url.replace(extname, '');
+      if (
+        (routeService?.extensions ?? DEFAULT_PAGE_EXTENSIONS).includes(extname)
+      ) {
+        url = url.replace(new RegExp(`\\${extname}$`), '');
```
JounQin, line 57: Align with

https://github.com/JounQin/rspress/blob/b951adf7f839523da754347ec4ed2bb9dc964e1c/packages/core/src/node/route/normalizeRoutePath.ts#L44

SoonIter (approved):

SoonIter:
thanks❤️️

PR #1994, opened 2025-03-23
refactor: improve header anchor logic and simplify URL validation

Simon-He95:
## Checklist

### DIFF
--- a/packages/core/src/node/mdx/remarkPlugins/checkDeadLink.ts
+++ b/packages/core/src/node/mdx/remarkPlugins/checkDeadLink.ts
@@ -65,7 +65,7 @@ export const remarkCheckDeadLinks: Plugin<
         return;
       }
 
-      if (!url.startsWith('http')) {
+      if (!url.startsWith('http://') && !url.startsWith('https://')) {
         const { routePath: normalizeUrl } = routeService.normalizeRoutePath(
           // fix: windows path
           url.split(path.sep).join('/')?.split('#')[0],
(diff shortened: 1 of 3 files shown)


packages/core/src/node/mdx/remarkPlugins/checkDeadLink.ts
```diff
@@ -65,7 +65,7 @@ export const remarkCheckDeadLinks: Plugin<
         return;
       }
 
-      if (!url.startsWith('http') && !url.startsWith('https')) {
+      if (!url.startsWith('http')) {
```
JounQin, line 68: Maybe we should check `http://` and `https://` instead?

packages/core/src/node/runtimeModule/globalStyles.ts
```diff
@@ -8,7 +8,7 @@ export async function globalStylesVMPlugin(context: FactoryContext) {
     await Promise.all(
```
JounQin, line 8: Then `Promise.all` is redundant.

Simon-He95: nice catch

packages/core/src/node/runtimeModule/globalStyles.ts
```diff
@@ -4,15 +4,12 @@ export async function globalStylesVMPlugin(context: FactoryContext) {
...
-          return `import ${JSON.stringify(source)};`;
-        }),
-    )
-  ).join('');
+  const moduleContent = [config?.globalStyles || '', ...globalStylesByPlugins]
+    .filter(source => source.length > 0)
+    .map(source => {
+      return `import ${JSON.stringify(source)};`;
+    })
+    .join('');
```
JounQin, line 12:
```suggestion
  const moduleContent = [config?.globalStyles || '', ...globalStylesByPlugins]
    .filter(Boolean)
    .map(source => `import ${JSON.stringify(source)};`)
    .join('');
```

SoonIter (approved):
thanks ❤️️

PR #1998, opened 2025-03-24
fix(mdxjs): add `version` and `lang` part when normalizing links

JounQin:
## Related Issue

related #1988

cc @SoonIter 

## Checklist

### DIFF
--- a/packages/core/src/node/mdx/options.ts
+++ b/packages/core/src/node/mdx/options.ts
@@ -52,6 +52,7 @@ export async function createMDXOptions(
         {
           cleanUrls,
           root: docDirectory,
+          routeService,
         },
       ],
       checkDeadLinks && [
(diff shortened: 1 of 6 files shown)


Timeless0911:
Frequent releases in the `1.x` branch may not be what we expect. If some issue fixes are not particularly serious, we hope to release them only in the `v2` version. Will you strongly rely on the `1.x` version or you can update to `v2` instead?

Timeless0911:
Got it, we can release `1.x` once a week on maybe Tuesday or Wednesday?

JounQin:
That's OK for non critical issues I think. For this one, we've workarounded by aligning `en/zh` contents, so it's not blocking us. So feel free to follow your normal schedule.

Timeless0911 (approved):

PR #2013, opened 2025-03-27
fix: use loose type of Rsbuild plugins

Timeless0911:
## Summary

use loose type of Rsbuild plugins

## Related Issue

https://github.com/web-infra-dev/rsbuild/blob/5f66734e1746ab1f8be990ce540fb0e68b19a83a/packages/core/src/types/plugin.ts#L225

## Checklist


Timeless0911:
1.x users can use `buildConfig.plugins` instead of `builderPlugins`.

PR #2018, opened 2025-03-28
feat(theme/toc)!: use dynamic toc generation

ktiays:
## Summary

this is an associated PR with https://github.com/web-infra-dev/rspress/pull/2101 of mdx refactoring

Otherwise, the TOC is wrong

and then we get more features

### 1. dynamic TOC generation even in user's tsx component

rspress-gif

image

https://lynxjs.org/guide/start/integrate-with-existing-apps.html#platform=ios

### 2. the `<Badge />` and more tsc component implementation in TOC

image

## Related Issue

close https://github.com/web-infra-dev/rspress/issues/1320

close https://github.com/web-infra-dev/rspress/discussions/1560

close https://github.com/web-infra-dev/rspress/discussions/1729

close https://github.com/web-infra-dev/rspress/pull/2036

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #2036, opened 2025-04-03
chore(theme-default/style): right asideContainer should exist even no headers

SoonIter:
## Summary

revert a style adjust to original behavior

image

#### before

image

#### after

image

## Checklist

### DIFF
--- a/packages/theme-default/src/layout/DocLayout/index.tsx
+++ b/packages/theme-default/src/layout/DocLayout/index.tsx
@@ -141,7 +141,7 @@ export function DocLayout(props: DocLayoutProps) {
               <ScrollToTop />
             </NoSSR>
           )}
-          {uiSwitch?.showAside && headers.length > 0 && (
+          {uiSwitch?.showAside && (
             <div
               className={styles.asideContainer}
               style={
@@ -150,9 +150,15 @@ export function DocLayout(props: DocLayoutProps) {
                   : { marginTop: 0, paddingTop: '32px' }
               }
             >
-              {beforeOutline}
-              <Aside headers={headers} outlineTitle={outlineTitle} />
-              {afterOutline}
+              {headers.length === 0 ? (
+                <></>
+              ) : (
+                <>
+                  {beforeOutline}
+                  <Aside headers={headers} outlineTitle={outlineTitle} />
+                  {afterOutline}
+                </>
+              )}
             </div>
           )}
         </div>


packages/theme-default/src/layout/DocLayout/index.tsx
```diff
@@ -150,9 +150,15 @@ export function DocLayout(props: DocLayoutProps) {
                   : { marginTop: 0, paddingTop: '32px' }
               }
             >
-              {beforeOutline}
-              <Aside headers={headers} outlineTitle={outlineTitle} />
-              {afterOutline}
+              {headers.length === 0 ? (
+                <></>
```
Timeless0911, line 154: If no headers, should `beforeOutline` and `afterOutline` take effect.

Timeless0911: https://github.com/web-infra-dev/rspress/discussions/1637

Timeless0911: image
 It should be a breaking change I think.

Timeless0911: @JounQin seems to resolve issue of overview page style?

JounQin:
Can we provide an option for this?

packages/theme-default/src/layout/DocLayout/index.tsx
JounQin: > @JounQin seems to resolve issue of overview page style?

Yes, I was.

JounQin:
```ts
headers.length === 0 && <>
//...
</>
```

Shoud be fine here.

JounQin: Can we add a new option for this behavior? We don't want to waste the window width.

SoonIter:
> Can we provide an option for this?

😂 sorry, 

after #2018, the TOC is only generated during CSR, so we need this UI placeholder in the right side for SSG. 

Otherwise, there will be a CLS

JounQin:
> 😂 sorry,
> 
> after #2018, the TOC is only generated during CSR, so we need this UI placeholder in the right side for SSG.
> 
> Otherwise, there will be a CLS

I think the `headers.length` could also be checked on SSG + CSR?

SoonIter:
> @SoonIter I think the `headers.length` could also be checked on SSG + CSR?

CSR: first screen(headers.length = 0) -> useEffect (headers.length = n)

SSG: html generation (headers.length = 0) -> first screen hydration(headers.length = 0) -> useEffect (headers.length = n)

So, I think this placeholder is necessary

To avoid CLS

in lg screen, I expect the first screen style to be like this, always with an AsideContainer for placeholder

image

in md screen, the first screen I expected would be

image




BTW, not the current context...

I find a bug when entering https://v2.rspress.dev/plugin/official-plugins/llms#usage in md screen

image


pending ...

PR #2055, opened 2025-04-11
feat(ssg)!: remove ssg fallback and add `ssg: false` hint

SoonIter:
## Summary

1. fallback to csr at compile time is too surprising for users, so just remove it 
2. add hint 

 no surprising error

## Checklist

### DIFF
--- a/packages/core/src/node/logger/hint.ts
+++ b/packages/core/src/node/logger/hint.ts
@@ -46,7 +46,7 @@ export async function hintThemeBreakingChange(customThemeDir: string) {
 
 export function hintSSGFailed() {
   logger.info(`[Rspress v2] \`ssg: true\` requires the source code to support SSR. If the code is not compatible to SSR, the build process will fail. You can try:
-    1. Fixed code to make it SSR-compatible.
+    1. Fix code to make it SSR-compatible.
     2. Set \`ssg: false\`, but the SSG feature will be lost.`);
 }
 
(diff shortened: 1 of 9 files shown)


packages/core/src/node/logger/hint.ts
```diff
@@ -43,3 +43,13 @@ export async function hintThemeBreakingChange(customThemeDir: string) {
     );
   }
 }
+
+export function hintSSGFailed() {
+  logger.info(`[Rspress v2] \`ssg: true\` requires the source code to support SSR. If the code is not compatible to SSR, the build process will fail. You can try:
+    1. Fixed code to make it SSR-compatible.
```
Timeless0911, line 49:
```suggestion
    1. Fix code to make it SSR-compatible.
```

Timeless0911 (dismissed):

packages/document/docs/en/api/config/config-basic.mdx
```diff
@@ -483,15 +483,15 @@ export default defineConfig({
...
-```ts title="rspress.config.ts"
-export default {
-  ssg: {
-    fallback: 'csr',
-  },
-};
-```
+SSG requires the source code to support SSR. If the code is not compatible to SSR, the build process will fail. You can try:
+
+1. Fixed code to make it SSR-compatible.
```
SoonIter, line 490:
```suggestion
1. Fix code to make it SSR-compatible.
```

Timeless0911 (approved):

PR #2065, opened 2025-04-14
feat: add printFileSize to SSG, and refactor renderPages to a rsbuildPlugin

SoonIter:
## Checklist


packages/core/src/node/ssg/rsbuildPluginSSG.ts
```diff
@@ -0,0 +1,97 @@
...
+                }
+              }
+
+              reject();
+            },
+          );
+        },
+      ).catch(() => {
+        const message =
+          'SSG requires `index.html`, but `index.html` is not emitted in web target';
```
Timeless0911, line 45:
```suggestion
          'SSG requires an `index.html` as entry, but this file is not emitted successfully in the web target.';
```

Timeless0911 (approved):

PR #2073, opened 2025-04-17
chore(deps)!: bump to react@19 and react-dom@19 by default

SoonIter:
## Summary

react@19 and react-dom@19

## Checklist


Timeless0911 (approved):

Timeless0911 (approved):

packages/plugin-preview/src/generate-entry.ts
```diff
@@ -25,10 +25,11 @@ export function generateEntry(
         `;
 
         const reactEntry = `
-        import { render } from 'react-dom';
+        import { createRoot } from 'react-dom/client';
```
Timeless0911, line 28: `react` is the `peerDependencies` of `@rspress/plugin-preview`, we should update the peer range if we support react 19.

colinaaa:
We are getting an error when upgrading to Rspress v2.0.0-beta.0:

```
website
└─┬ @rspress/runtime 2.0.0-beta.0
  └─┬ react-helmet-async 2.0.5
    └── ✕ unmet peer react@"^16.6.0 || ^17.0.0 || ^18.0.0": found 19.1.0 in @rspress/runtime
hint: If you don't want pnpm to fail on peer dependency issues, add "strict-peer-dependencies=false" to an .npmrc file at the root of your project.
```
> https://github.com/lynx-family/lynx-stack/pull/570#issuecomment-2818162296

Should we just ignore the error? Or could Rspress handle this (e.g.: bundle `react-helmet-async` to avoid using it in `dependencies`)?

Ref: https://github.com/staylor/react-helmet-async/issues/239

PR #2079, opened 2025-04-18
feat: add `_meta.json` json schema in `rspress/meta-json-schema.json`

SoonIter:
## Summary

1. <s>rename top level _meta.json => nav.json</s>

2. add _meta.json json schema type hint

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,6 @@ e2e/fixtures/plugin-typedoc/multi/doc/api
 
 # ignore typed css module
 *.scss.d.ts
+
+# ignore generated schema.json
+meta-json-schema.json
--- a/.prettierignore
+++ b/.prettierignore
@@ -5,6 +5,7 @@ doc_build
 
 # ignore all JS/TS files, use Biome
 **/*.js
+**/*.mjs
 **/*.ts
 **/*.jsx
 **/*.tsx
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -13,6 +13,12 @@
   "files.exclude": {
     "**/.DS_Store": true
   },
+  "json.schemas": [
+    {
+      "fileMatch": ["**/_meta.json"],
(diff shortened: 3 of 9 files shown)


CoffeeChaton:
In this pr, the doc doesn't seem to mention `meta-json-schema.json`.
Therefore, I was wondering if there are any plans to make `meta-json-schema.json` publicly available?

I'm hoping to reduce how often I need to leave the IDE, and allow me to use it as follows for in-editor validation and autocompletion:

```jsonc
// .vscode/settings.json
{
  "json.schemas": [
    {
      "fileMatch": ["**/_meta.json"],
      "url": "./node_modules/rspress/xxxPath/meta-json-schema.json"
      // or "meta-json-schema-base.json"
    }
  ],
}
```

SoonIter:
> In this pr, the doc doesn't seem to mention `meta-json-schema.json`. Therefore, I was wondering if there are any plans to make `meta-json-schema.json` publicly available?

Yeah, of course, it is one of the public features in v2.0, 

v2.0 is in prerelease version, and document is not fully covered now

CoffeeChaton:
Thank you.

Timeless0911:
We need to carefully consider whether should we add a new `_nav.json`.

JounQin:
I personally love splitting `_nav.json` with the old `_meta.json`, because in our case, there is only on sidebar, the root `_meta.json` is used for sidebar but we'd expect it as navbar instead.

SoonIter:
> I personally love splitting `_nav.json` with the old `_meta.json`, because in our case, there is only one sidebar, the root `_meta.json` is used for sidebar but we'd expect it as navbar instead.

yeah,  splitting `_nav.json` with the old `_meta.json` brings that this case can be handled easier than before, 

but there have already been too many breaking changes here... so I decided to give up this change

Timeless0911 (dismissed):

Timeless0911 (approved):

PR #2093, opened 2025-04-24
feat(plugin-llms): add include option and only generate for default lang

SoonIter:
## Summary

feat(plugin-llms): add include option and only generate for default lang

## Checklist


chenjiahan (approved):
LGTM 👍

PR #2125, opened 2025-05-07
feat: add `localeRedirect: 'only-default-lang'` support

JounQin:
## Summary

If you set it to `only-default-lang`, the user will only be redirected when visiting the default locale.

## Checklist


SoonIter (approved):
thanks~

PR #2139, opened 2025-05-09
feat: update Rsbuild 1.3.18 and set caller name

chenjiahan:
## Summary

Update Rsbuild to 1.3.18 and set the `callerName` option of `createRsbuild`, which can be accessed by Rsbuild plugins through `context.callerName`, and execute different logic based on this identifier.

## Related Links

- https://rsbuild.dev/api/javascript-api/core#specify-caller-name
- https://github.com/web-infra-dev/rsbuild/releases/tag/v1.3.18

## Checklist


Timeless0911 (approved):

JounQin:
I'm wondering should we use `~` range for `@rsbuild/core` just like others?

chenjiahan:
@JounQin Nice catch, I forgot why we pinned the @rsbuild/core version in the past (maybe we encountered a bug at the time), now unpinning is a better option. I have updated it in this PR.

JounQin:
Actually as a stable version (`@rsbuild/core` v1+), we should mostly use `^` range at least in the future?

chenjiahan:
@JounQin We prefer to use `~` to avoid introducing Rspack's experimental features and SWC's Wasm plugins breaking changes to Rspress.  see https://rspack.dev/blog/announcing-1-0#does-rspack-follow-semantic-versioning for details.

PR #2148, opened 2025-05-12
chore(plugin-shiki): add all builtin-langs in shiki

SoonIter:
## Summary

https://github.com/SoonIter/shiki/blob/f649d11c14aa31e4487a87e2cb0b547ce4f00d04/packages/rehype/src/index.ts#L30

drop the default value of langs in plugin-shiki, so langs option can be needless

## Checklist

### DIFF
--- a/packages/document/rspress.config.ts
+++ b/packages/document/rspress.config.ts
@@ -26,7 +26,6 @@ export default defineConfig({
   markdown: {
     checkDeadLinks: true,
     shiki: {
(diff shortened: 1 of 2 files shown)


Timeless0911 (approved):

JounQin:
No any downside effects?

SoonIter:
> No any downside effects?

Rspress does highlight during compile time, so users do not need to register the language.

PR #2149, opened 2025-05-12
feat(core)!: routeService ignore "_" prefix files by default

SoonIter:
## Summary

Docusaurus has a Partial mdx convention, starting with '_', mainly used to exclude document fragments in convention-based routing.

`import PartialExample from './_partial.mdx'`

to replace

```ts
// rspress.config.ts
route: {
   exclude: ['**/shared/**']
}
```

## Related Issue

https://docusaurus.io/docs/markdown-features/react#importing-markdown

## Checklist


JounQin:
Would `!**/_*` restore the previous behavior?

SoonIter:
> Would `!**/_*` restore the previous behavior?

what previous behavior?

JounQin:
> what previous behavior?

`_foo.mdx` won't be ignored, it'll be used as `_foo` route.

PR #2159, opened 2025-05-14
docs(tailwind-4): update how to use tailwind 4.0

Gao-pw:
## Summary
updated postcss usage on using tailwind@4

## Related Issue

#2158 

## Checklist


Gao-pw:
oops, I think v2 may not be suitable at present, so I will close it first

PR #2174, opened 2025-05-16
perf: remove React 17 compatibility code

chenjiahan:
## Summary

This PR removes React 17 compatibility code from Rspress.

Rspress 2.0 uses React 19 by default, and the minimum supported React version is 18, which means React 17 will no longer be supported.

## Related Issue

- https://github.com/web-infra-dev/rspress/discussions/1891#discussioncomment-12475395

## Checklist


SoonIter (approved):
thanks~

Issue #1912, opened 2025-03-06
[Feature]: Save user-preferred language in storage


JounQin:
### What problem does this feature solve?

When the user visits https://rspress.dev at the first time, it will be redirected to https://rspress.dev/zh/ automatically, but when he visits the second time, no redirect follows, it's very confusing.

### What does the proposed API look like?

Save user selected language into `localStorage`, if the current language doesn't match, redirect to the preferred language automatically, the preferred language will be changed on toggling language manually by the user.

SoonIter:
feel free to contribute these hooks, like `useLang`

IMO, these business UI logics are very common, even switching to a community widely used hook library is under my expectation

HyaCiovo:
If a user has explicitly selected a preferred language before, it makes sense to save their choice in storage.
But users may intentionally want to visit a specific language version (e.g., via direct link .../zh/) , and they will be blocked from doing so due to the redirect? (I'm not sure🤔)
This can be confusingw — especially for sharing links — because the system may override the URL without clear feedback.

JounQin:
@HyaCiovo It could follow https://rspress.dev/api/config/config-theme#localeredirect

PR #2187, opened 2025-05-22
feat(theme): Save user preferred language and optimize navigation (#1912)

HyaCiovo:
## Summary
- Add `saveUserPreferredLanguage` property to Config interface to control whether to save user switched language
- Update navigation component to add click event handling for language switching
- Add description for `saveUserPreferredLanguage` configuration

image

## Related Issue
- #1912

## Checklist

### DIFF
--- a/packages/document/docs/en/api/config/config-theme.mdx
+++ b/packages/document/docs/en/api/config/config-theme.mdx
@@ -767,3 +767,20 @@ export default defineConfig({
   },
 });
 ```
+
+## saveUserPreferredLanguage
+
+- Type: `'auto' | 'never' | 'only-home-page'`
+- Default: `'auto'`
+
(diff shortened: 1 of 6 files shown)


SoonIter:
Thanks again ❤

HyaCiovo:
During my local testing, I found that the redirection timing was really too late. I agree that implementing it in inline HTML scripts should to be a better option, offering earlier execution and greater convenience for unit testing.👍

As I happened to notice that the `useRedirect4FirstVisit` also has logic that relies on variables stored in `localStorage` for redirection, I thought  that a similar hook could be used to achieve my goal at that time.
Additionally, there is a lot of redirection logic in the current `theme` package, and I was really troubled by the interaction of various logics during local development. Your view is indeed correct. Such logic really shouldn't be placed in the `theme` dependency.

Issue #2126, opened 2025-05-07
[Feature]: Runtime Component <ReactMarkdown /> for mdx parsing and <CodeBlockRuntime /> for code block


SoonIter:
### What problem does this feature solve?

https://github.com/web-infra-dev/rspress/pull/2122#issuecomment-2854256408

> The runtime component will increase the bundle size and will not be recorded in static data such as searchIndex.
> it is out of the control of Rspress literally, they are advanced usages
> So we can provide a Component with md syntax parsing at runtime just like `<ReactMarkdown />` and `<CodeBlockRuntime />` and clarify the difference from Rspress compile time
> https://fumadocs.dev/docs/ui/components/dynamic-codeblock

### What does the proposed API look like?

1. mdx parsing runtime component

a runtime component like react-markdown integrated with default remark plugins

https://github.com/web-infra-dev/rspress/blob/0a75ac2be9e54ee58a1555a0179e28efefc43f06/packages/plugin-api-docgen/static/global-components/API.tsx

2. code block parsing runtime component

```ts
<CodeBlockRuntime />
```

SoonIter:
We encourage users to use code blocks at compile time ` ```tsx ` as much as possible, not the runtime ones; this can only be considered an escape hatch.

We'll provide the documentation on the limitations of CodeBlockRuntime

PR #2190, opened 2025-05-22
feat: add new `CodeBlockRuntime` component

JounQin:
## Summary

add new `CodeBlockRuntime` component

## Related Issue

related #2126

## Checklist


SoonIter (commented):
could you add some test cases?

SoonIter:
This shiki dependency will strongly depend on esm's treeshaking to be clear.

So after merging this, we will observe the increase in bundle size in the coming days.

JounQin:
> could you add some test cases?

I don't see test cases for other built-in components so I didn't, I'll add an e2e test case soon later.

JounQin:
> This shiki dependency will strongly depend on esm's treeshaking to be clear.
> 
> So after merging this, we will observe the increase in bundle size in the coming days.

So, https://github.com/un-ts/mini-shiki should be considered, https://github.com/un-ts/mini-shiki/issues/3.

SoonIter:
> I don't see test cases for other built-in components so I didn't, I'll add an e2e test case soon later.

I think it should actually be placed in `@rspress/runtime`, because it is a feature and unrelated to style.

For the convenience of users to import `import { CodeblockRuntime } from '@theme'`, I think it's also okay to put it in `@rspress/theme-default`.

For style, we can skip writing tests because it's really difficult to write the style behavior. lol

So, you can write the test into `plugin-shiki.test.ts`, test it together with other shikis, only test the simplest code blocks is acceptable, trust me, this is a very simple verification.

JounQin:
@SoonIter Just added an e2e test case. :)

SoonIter:
it looks charming❤

SoonIter (approved):

PR #2191, opened 2025-05-22
chore(deps): remove htmr dep and migrate htmr to `dangerouslySetInnerHTML`

JounQin:
## Summary

remove htmr dep

## Related Issue

close https://github.com/web-infra-dev/rspress/issues/2232

## Checklist

### DIFF
--- a/e2e/tests/inline-markdown.test.ts
+++ b/e2e/tests/inline-markdown.test.ts
@@ -91,11 +91,11 @@ test.describe('Inline markdown test', async () => {
       [
         'Class: Component&lt;P, S&gt;',
         'Class: Component&lt;P, S&gt;',
-        '<strong>bold</strong>',
-        '<em>emphasis</em>',
-        '<code>&lt;foo&gt;</code>',
+        '<span><strong>bold</strong></span>',
+        '<span><em>emphasis</em></span>',
+        '<span><code>&lt;foo&gt;</code></span>',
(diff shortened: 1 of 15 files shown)


SoonIter:
This is not a lossless change because there is an additional `span` level.

Could we use some of the code here of dom api to remove this `span` ?

https://github.com/pveyes/htmr/blob/master/src/index.browser.ts

JounQin:
> This is not a lossless change because there is an additional `span` level.
> 
> Could we use some of the code here of dom api to remove this `span` ?
> 
> [pveyes/htmr@`master`/src/index.browser.ts](https://github.com/pveyes/htmr/blob/master/src/index.browser.ts?rgh-link-date=2025-05-23T08%3A44%3A24Z)

Then why not just continue using `htmr`?

JounQin:
@SoonIter In case that we've already included rehype related tools due to `CodeBlockRuntime`/`shiki`, I think we should use `rehype-parse` + `hast-to-jsx-runtime` instead, WDYT?

SoonIter:
> @SoonIter In case that we've already included rehype related tools due to `CodeBlockRuntime`/`shiki`, I think we should use `rehype-parse` + `hast-to-jsx-runtime` instead, WDYT?

For performance, when we do not need to perform modification on hast, I prefer using `dangerouslySetInnerHTML` directly.

Issue #2186, opened 2025-05-21
[Bug]: pluginLlms


dragooncjw:
### Version

```sh
system: macOS
node version: v18.20.4
```

### Details

rspress version: "rspress": "^1.38.0",

报错如下图：
![Image](https://github.com/user-attachments/assets/df0e92f1-d3e0-431d-ace3-648b7382dce6)

此外，plugin-llms 插件似乎 document 路径是错误的：https://rspress.dev/plugin/official-plugins/llms 这个路径 404，是因为还在建设中吗？

### Reproduce link

https://codesandbox.io/p/devbox/laughing-water-5gkg3r?file=%2Fpackage.json%3A12%2C6-12%2C26

### Reproduce Steps

1. "rspress": "^1.38.0",
2. "@rspress/plugin-llms": "^2.0.0-alpha.8",
3. in rspress.config.ts, config `plugins: [pluginLlms()],`

Timeless0911:
Current docs of Rspress v2 is under https://v2.rspress.dev/plugin/official-plugins/llms.

And when the v2 docs become more stable, we will change it to rspress.dev and make v1 docs to be v1.rspress.dev

SoonIter:
Version needs to be consistent

```json
"devDependencies": {
   "rspress": "2.0.0-beta.10",
   "@rspress/plugin-llms": "2.0.0-beta.10",
}
```

PR #2193, opened 2025-05-23
chore(plugin-llms): support config targets

dragooncjw:
## Summary
Support custom targets of plugin-llms

## Related Issue
https://github.com/web-infra-dev/rspress/issues/2186

## Checklist


Timeless0911:
Documentation should be updated if you add a new option. And there exists type error during build.

SoonIter:
I think your question is about how to use plugin-llms when `ssg: false`,

Currently, llms can only be used with `ssg: true`.
When the user sets `ssg: false`, we can set targets to `['web']`.

SoonIter:
suggestion: flowgram.ai can try using `ssg: true` to make the SEO work

PR #2205, opened 2025-05-26
fix: `children` should be unavailable for `CodeBlockRuntime`

JounQin:
## Summary

image

## Related Issue

related #2190

## Checklist


JounQin:
> could title be optional?

It's already optional.

> And we also need a shikiOptions to pass shiki options

OK, I'll raise another PR for it.

PR #2207, opened 2025-05-27
Revert "feat: enable `rehype-raw` plugin by default (#2199)"

SoonIter:
This reverts commit fb2ff047ecde023f03d71721bec985a50d0763ea.

## Summary

````
```js title="test.js"
console.log('Hello CodeBlock!');
```
````

title does not work after adding `rehype-raw`

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #2213, opened 2025-05-28
feat: enable `rehype-raw` plugin by default

JounQin:
## Summary

enable `rehype-raw` plugin by default by adding a new tiny `rehypeRawFix` plugin at the same time

It's just working as expected now.

image

## Related Issue

related #2207

## Checklist

### DIFF
--- a/packages/core/src/node/mdx/options.ts
+++ b/packages/core/src/node/mdx/options.ts
@@ -86,9 +86,13 @@ export async function createMDXOptions(
       ],
       ...rehypePluginsFromConfig,
       ...rehypePluginsFromPlugins,
+      // 1. why adding rehype-raw?
       // This is what permits to embed HTML elements with format 'md'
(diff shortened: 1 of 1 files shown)


packages/core/src/node/mdx/options.ts
```diff
@@ -83,6 +86,10 @@ export async function createMDXOptions(
       ],
       ...rehypePluginsFromConfig,
       ...rehypePluginsFromPlugins,
-    ],
+      // This is what permits to embed HTML elements with format 'md'
+      // See https://github.com/facebook/docusaurus/pull/8960
+      // See https://github.com/mdx-js/mdx/pull/2295#issuecomment-1540085960
+      format === 'md' && [rehypeRaw, { passThrough: nodeTypes }],
```
JounQin, line 92: As I mentioned, would this work for 3rd-party plugins which expects `rehype-raw` to be used first?

PR #2225, opened 2025-05-30
docs: `facet` should be `lang:` instead of `language:`

JounQin:
## Summary

`facet` should be `lang:` instead of `language:`

## Related Issue

N/A

## Checklist


JounQin:
replaced by #2226

PR #2226, opened 2025-05-30
docs: upgrade the algolia template

SoonIter:
## Summary

https://docsearch.algolia.com/docs/templates#rspress-template

## Checklist


packages/document/docs/en/plugin/official-plugins/algolia.mdx
```diff
@@ -202,11 +202,16 @@ new Crawler({
       indexName: 'doc_search_rspress_pages',
```
JounQin, line 202: This should be different for every project?

SoonIter: `Here is an example config based on what this site uses:`

so I think it is okay

PR #2237, opened 2025-06-04
docs: use escape character to bypass HTML render

fi3ework:
## Summary

Seems like v2's doc has fixed this issue, this is just a quick fix without investigating why it happens (maybe a renderer bug).

https://rspress.dev/plugin/official-plugins/preview#internal-components

before:

image

after:

image

## Checklist


SoonIter (approved):
lgtm

PR #2241, opened 2025-06-05
chore(deps): upgrade @types/react and @types/react-dom to 19

SoonIter:
## Summary

chore(deps): upgrade @types/react and @types/react-dom to 19

## Checklist


Timeless0911 (dismissed):

packages/theme-default/src/layout/DocLayout/docComponents/pre.tsx
```diff
@@ -97,7 +97,9 @@ export function PreWithCodeButtonGroup({
   const preElementRef = useRef<HTMLPreElement>(null);
 
   const renderChild = (child: React.ReactElement) => {
```
JounQin: Why not:

```suggestion
  const renderChild = (child: React.ReactElement<{ className?: string }>) => {
```

packages/theme-default/src/layout/DocLayout/docComponents/pre.tsx
```diff
@@ -120,5 +119,7 @@ export function PreWithCodeButtonGroup({
 
   if (!isValidElement(children)) return null;
```
JounQin:
```suggestion
  if (!isValidElement<{ className?: string }>(children)) return null;
```

Not need to use `as`

Timeless0911 (approved):

PR #2267, opened 2025-06-11
feat: support async entry

2heal1:
## Summary
MF will replace the original entry with async entry like below: 

```js
// original entry - ssrEntry.js
export {reder,routes};

// after access mf - bootstrap.js
const entry = import('ssrEntry.js');
const render = entry.then({render}=> (render));
const routes = entry.then({routes}=> (routes));

export {render, router}
```

And the ssrExport.render will be Promise.resolve(render), so need to add `await`

## Checklist

### DIFF
--- a/.changeset/vast-words-enter.md
+++ b/.changeset/vast-words-enter.md
@@ -1,0 +1,5 @@
+---
+'@rspress/core': patch
+---
+
+feat: support async entry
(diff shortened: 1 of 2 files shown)


SoonIter:
We can later add e2e tests for `pluginModuleFederation` to ensure this feature won't be broken

PR #2286, opened 2025-06-16
fix(lazyCompilation): use `{ entries: false }` for better DX

SoonIter:
## Summary

Rspress actually has only one entry, so there is no need for lazy loading. It is the necessary path for rendering.

Currently, we have adopted lazyLoading, which will only load CSS when the developer first opens the page, resulting in a poor experience (large CLS).

This PR may lead to degradation in benchmark data, but these are necessary.

### before

image

### after

This result is actually faster than before because some work has already been done during the time the user opens the browser.

image

## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -173,7 +173,13 @@ async function createInternalBuildConfig(
       },
     },
     dev: {
-      lazyCompilation: process.env.RSPRESS_LAZY_COMPILATION !== 'false', // This is an escape hatch for playwright test, playwright does not support lazyCompilation
+      // This is an escape hatch for playwright test, playwright does not support lazyCompilation
+      lazyCompilation:
+        process.env.RSPRESS_LAZY_COMPILATION === 'false'
+          ? false
+          : {
+              entries: false,
+            },
       progressBar: false,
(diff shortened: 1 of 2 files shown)


chenjiahan (commented):
Maybe this should be the default behavior of Rsbuild's `dev.lazyCompilation: true`? (set `entries: false` when there is only one entry)

SoonIter:
> Maybe this should be the default behavior of Rsbuild's `dev.lazyCompilation: true`? (set `entries: false` when there is only one entry)

1. For MPA, `entries: true` is very useful, (maybe the number of entries is larger than 5)

2. For SPA, `entries: false` is acceptable

@JSerFeng what's your opinion

SoonIter:
> Maybe this should be the default behavior of Rsbuild's `dev.lazyCompilation: true`? (set `entries: false` when there is only one entry)

LGTM

```ts
{
  entries: Object.keys(source.entries).length > 1
}
```

chenjiahan:
Rsbuild PR: https://github.com/web-infra-dev/rsbuild/pull/5431

JSerFeng:
> For SPA, entries: false is acceptable

I agree, `entries: false` is almost useless for SPA

PR #2291, opened 2025-06-17
refactor!: remove useless "id" field in `PageIndexInfo`

SoonIter:
## Summary

1. refactor!: remove id field in PageIndexInfo

routePath should be used as id, so id is no need to be maintained

2. refactor code style: use class singleton `static async create()`

## Checklist


JounQin:
https://github.com/jl917/rspress-plugin-sitemap/blob/04e2a9daa452f838d0d7fc4b317c7ea7aad807b5/src/index.ts#L61 is affected, cc @jl917

PR #2306, opened 2025-06-20
chore(infra)!: move "@rspress/plugin-auto-nav-sidebar" to @rspress/core

SoonIter:
## Summary

@rspress/plugin-auto-nav-sidebar is deprectaed

## Checklist


packages/core/src/node/auto-nav-sidebar/index.ts
```diff
@@ -78,7 +78,7 @@ export function pluginAutoNavSidebar(): RspressPlugin {
         if (hasLang) {
```
JounQin, line 78: I think this is a false positive personally, why the user can not enforce the `lang` as `zh` (default is `en`)?

SoonIter: i do not know, just move this pkg, I will continue to modify this part.

Issue #1677, opened 2024-12-18
[Feature]: dynamic base href support


JounQin:
### What problem does this feature solve?

We want to deploy our documents into multiple sites, for siteA its base could be `/docs`, but for siteB its base could be `/platform-docs`, for now we have to build them multiple times separately.

> [!WARNING]  
> This feature could break SSG

### What does the proposed API look like?

Proposal:

1. check `typeof __RSPRESS_BASE_HREF__`
2. check [`base#href`](https://developer.mozilla.org/zh-CN/docs/Web/HTML/Element/base#href)
3. provide `function` support for runtime check

Timeless0911:
I think building multiple times is a reasonable solution, and all other solutions will increase the complexity.

SoonIter:
after https://github.com/web-infra-dev/rspress/pull/2322, it is possible to implement

However, there is still a certain risk regarding accuracy if some compile-time behaviors within core or plugins depend on the `base` configuration.

PR #2322, opened 2025-06-25
refactor(runtime)!: use `basename` feature of react-router

SoonIter:
## Summary

https://github.com/facebook/docusaurus/blob/45065e8d2b5831117b8d69fec1be28f5520cf105/packages/docusaurus/src/client/exports/Link.tsx#L78

I 'm done with `withBase` and `removeBase` and `withoutBase` `addLeadingSlash` `addTrailingSlash`

it is silly and wasting my life

## Related Issue

close https://github.com/web-infra-dev/rspress/pull/2321

## Checklist


packages/core/src/node/utils/normalizePath.ts
```diff
@@ -41,9 +41,9 @@ function absolutePathToRoutePrefix(
     ),
   );
   const routeService = RouteService.getInstance();
-  const [basePrefix, versionPrefix, langPrefix] =
+  const [_, versionPrefix, langPrefix] =
```
JounQin:
```suggestion
  const [versionPrefix, langPrefix] =
```

If I didn't mis-read, there is no `base` now?

packages/core/src/node/utils/normalizePath.ts
```diff
@@ -41,9 +41,9 @@ function absolutePathToRoutePrefix(
     ),
   );
   const routeService = RouteService.getInstance();
-  const [basePrefix, versionPrefix, langPrefix] =
+  const [versionPrefix, langPrefix, _] =
```
JounQin:
```suggestion
  const [versionPrefix, langPrefix] =
```

I don't think `_` is needed here?

packages/core/src/node/runtimeModule/runtimeConfig.ts
```diff
@@ -0,0 +1,17 @@
...
+ */
+export const runtimeConfigVMPlugin: VirtualModulePlugin = context => {
+  return {
+    [RuntimeModuleID.RuntimeConfig]: () => {
+      const { config } = context;
+
+      const { base } = config;
+
+      return `
+export const base = ${JSON.stringify(base)};`;
```
JounQin:
```suggestion
      return `export const base = ${JSON.stringify(base)};`;
```

It's weird to have starting blanks without ending blanks?

JounQin:
Could this PR make #1677 possible? Example virtual runtime module:

```js
`export const base = typeof __RSPRESS_BASE_HREF__ === 'undefined' ? ${JSON.stringify(base)} : __RSPRESS_BASE_HREF__`
```

SoonIter:
> Could this PR make #1677 possible? Example virtual runtime module:
> 
> ```js
> `export const base = typeof __RSPRESS_BASE_HREF__ === 'undefined' ? ${JSON.stringify(base)} : __RSPRESS_BASE_HREF__`
> ```


It's really similar to publicPath configuration in webpack

There are also some compile-time stages that read the base configuration like plugins, and we can't guarantee that they are correct either if we only modify them at runtime.

SoonIter:
> There are also some compile-time stages that read the base configuration like plugins, and we can't guarantee that they are correct either if we only modify them at runtime.

For the core package, I can refactor it to no rely on the `base`. In the future, it may be possible to implement, but some rspress plugins may depend on base configuration, and this feature carries a certain risk of accuracy.

SoonIter:
close https://github.com/web-infra-dev/rspress/pull/2321

PR #2343, opened 2025-07-01
feat(theme-default): support binary execution commands in `PackageManagerTabs` component

artus9033:
## Summary

This PR adds a new prop to the `PackageManagerTabs` component for binary execution (non-install) commands. This covers a use case when the `PackageManagerTabs` component would be used to present documentation on how to execute a binary tool provided by an installed package, e.g. `npx react-native ...`.

The current design of this component either accepts a string that is prefixed with the package manager binary, with conditional postprocessing of flags. This does work for yarn, pnpm & bun, but does not suit npm-installed binaries, where `npx` should be used instead. A possible workaround is to pass a full object separately specifying the full command for each of the package managers, in which case the auto-prefixing is not applied (such as in https://github.com/callstackincubator/react-native-legal/pull/53). In that case, the commands need to explicitly include the executable, which is suboptimal and aggravates code readability.

To overcome this, the proposed PR introduces an optional `exec` boolean flag applicable only if the `command` prop is a string, in which case `exec = true` changes the `npm` prefix to `npx` for the npm package manager tab.

## Related Issue

N/A

## Checklist

- [x] Tests updated ~~(or not required)~~.
- [x] Documentation updated ~~(or not required)~~.

### DIFF
--- a/packages/document/docs/en/fragments/builtin-components.mdx
+++ b/packages/document/docs/en/fragments/builtin-components.mdx
@@ -322,27 +322,39 @@ import { PackageManagerTabs } from '@theme';
 The types of props included are as follows:
 
 ```ts
-interface PackageManagerTabsProps {
-  command:
-    | string
-    | {
+type PackageManagerTabProps = (
+  | {
+      command: string;
+      /**
+       * If true, the command will be interpreted as a shell command and prefixed with npx for npm,
+       * or the package manager binary for others.
+       */
+      exec?: boolean;
+    }
+  | {
+      command: {
         // Used to set commands for different package managers
         npm?: string;
         yarn?: string;
         pnpm?: string;
         bun?: string;
       };
+      exec?: never;
+    }
+) &
   // Used to set additional tabs
-  additionalTabs: {
-    // Used to set additional package managers
-    tool: string;
-    // Used to set the icon of the additional package manager
-    icon?: React.ReactNode;
-  }[];
-}
+  {
+    additionalTabs?: {
+      // Used to set additional package managers
+      tool: string;
+      // Used to set the icon of the additional package manager
+      icon?: ReactNode;
+    }[];
+  };
 ```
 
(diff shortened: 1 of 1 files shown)


packages/theme-default/src/components/PackageManagerTabs/index.tsx
```diff
@@ -73,7 +84,7 @@ export function PackageManagerTabs({
   // Init Command
   if (typeof command === 'string') {
     commandInfo = {
-      npm: `npm ${command}`,
+      npm: `${exec ? 'npx' : 'npm'} ${command}`,
       yarn: `yarn ${command}`,
```
SoonIter, line 88: Is this only for handling npm?  Can yarn, pnpm, and bun work?

chatgpt answers me

```sh
# npm
npx cowsay "Hello World"

# yarn 1.x
yarn cowsay "Hello World"  # only when yarn global add cowsay

# yarn 2+
yarn dlx cowsay "Hello World"

# pnpm
pnpm dlx cowsay "Hello World"

# bun
bunx cowsay "Hello World"
```

JounQin: Without `dlx`/`bunx` is also unnecessary.

`yarn cowsay "Hello World"  # only when yarn global add cowsay` comment is wrong.

```sh
# npm
npx cowsay "Hello World"

# yarn
yarn cowsay "Hello World"

# pnpm
pnpm cowsay "Hello World"

# bun
bun cowsay "Hello World"
```

`yarn/pnpm dlx` and `bunx` only make sense when there is a same-named script defined.

SoonIter:
Thanks for your contribution, could you please sign the CLA?

artus9033:
All done @SoonIter  ✅

Timeless0911:
Can anyone translate this doc to Chinese and send a PR?

PR #2348, opened 2025-07-03
fix(remarkNormalizeLink)!: support relative path [subfolder](subfolder) which is equal to [subfolder](./subfolder)

SoonIter:
## Summary

1. support relative path `[subfolder](subfolder)` equal to `[subfolder](./subfolder)`

image

**Before**

`[](installation)` and `[](installation.mdx)` points to `/installation` which is strange ❌

image

**After**

`[](installation)` and `[](installation.mdx)` points to `/guide/api/installation.html` ✅

2. fix(RouteService): `/zh` should be `/zh/`

## Checklist


packages/document/docs/en/api/config/config-basic.mdx
```diff
@@ -409,7 +409,7 @@ export default defineConfig({
 
 :::note
 
-We recommend using [addPages hook](plugin/system/plugin-api.html#addpages) in a custom Rspress plugin to add some additional files to the route, so that the page route and file path/content can be specified more flexibly and reasonably.
+We recommend using [addPages hook](/plugin/system/plugin-api.html#addpages) in a custom Rspress plugin to add some additional files to the route, so that the page route and file path/content can be specified more flexibly and reasonably.
```
JounQin, line 412: So, this is another breaking change.

SoonIter: yes, this is an ancient bug that has caused many incorrect uses.

SoonIter:
https://github.com/web-infra-dev/rspress/discussions/1891#discussioncomment-13659861

PR #2349, opened 2025-07-04
feat(rsbuild)!: enable persistent cache by default via `performance.buildCache`

SoonIter:
## Summary

feat(rsbuild)!: enable persistent cache defaultly via `performance.buildCache` push

### rspress build

image

image

### rspress dev

image

## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -229,7 +229,9 @@ async function createInternalBuildConfig(
       },
     },
     performance: {
-      buildCache: true,
+      buildCache: {
+        buildDependencies: [pluginDriver.getConfigFilePath()],
+      },
       chunkSplit: {
         override: {
           cacheGroups: {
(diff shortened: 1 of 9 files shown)


packages/core/src/node/initRsbuild.ts
```diff
@@ -229,6 +229,9 @@ async function createInternalBuildConfig(
       },
     },
     performance: {
+      buildCache: {
+        buildDependencies: [pluginDriver.getConfigFilePath()],
```
chenjiahan, line 233: Is Rsbuild's default buildDependencies not work as expected?

https://rsbuild.rs/config/performance/build-cache#builddependencies

Image

SoonIter: still investigating

SoonIter: At present, the answer is  "yes"

after adding buildDependencies `rspress.config.ts`, CI is green

Failed case before

image

chenjiahan: 👌

chenjiahan (approved):

PR #2360, opened 2025-07-09
feat!: expose type `RspressPlugin` from "rspress/core", "@rspress/shared" which is a private package should not be installed by users

SoonIter:
## Summary

We do not want users to use our private package, only `"rspress": "2.0.0"` package is needed

before

```ts
import type { RspressPlugin } from '@rspress/shared';
```

after

```ts
import type { RspressPlugin } from 'rspress/core';
```

## Checklist


packages/core/src/node/last-updated/index.ts
```diff
@@ -1,5 +1,3 @@
-/// <reference path="../index.d.ts" />
```
JounQin: Why it's removed and doesn't work anymore?

SoonIter: yes, not work now
image

JounQin (approved):

Timeless0911 (approved):

PR #2363, opened 2025-07-10
fix: support mutations from child nodes for dynamic TOC

### LINKED ISSUE #2346: It seems this only supports static TOC on the first render, for real dynamic TOC generation, `MutationObserver` could be considered.
It seems this only supports static TOC on the first render, for real dynamic TOC generation, `MutationObserver` could be considered.

Use case: A dynamic component which fetches data on runtime like https://github.com/alauda/doom/blob/44d293aa9e458b9521dfce9f096442222db15dbb/src/runtime/hooks/useSiteOverrides.ts#L125

_Originally posted by @JounQin in https://github.com/web-infra-dev/rspress/issues/2018#issuecomment-3031309351_

JounQin:
## Summary

support mutations from child nodes for dynamic TOC

also refactor to use `for of` to reduce loop execution

## Related Issue

close #2346

## Checklist


SoonIter (approved):

SoonIter:
thanks❤

PR #2370, opened 2025-07-11
test: add missing non-null assertion to dynamic-toc test

Marukome0743:
## Summary

Tests:
- Apply ‘!’ operator to textContent access in dynamic-toc tests to satisfy non-null assertions

## Related Issue or Pull Request

Follow up #2363

## Checklist

### DIFF
--- a/e2e/fixtures/dynamic-toc/index.test.ts
+++ b/e2e/fixtures/dynamic-toc/index.test.ts
@@ -25,21 +25,21 @@ test.describe('dynamic toc', async () => {
     });
 
     let h2 = await page.$('h2');
-    let text = await page.evaluate(h2 => h2?.textContent.trim(), h2);
+    let text = await page.evaluate(h2 => h2?.textContent!.trim(), h2);
     expect(text).toBe('#Term');
 
     let toc = await page.$('.aside-link');
-    let tocText = await page.evaluate(toc => toc?.textContent.trim(), toc);
+    let tocText = await page.evaluate(toc => toc?.textContent!.trim(), toc);
     expect(tocText).toBe('Term');
 
     await setTimeout(1000); // Wait for dynamic TOC to update
 
     h2 = await page.$('h2');
-    text = await page.evaluate(h2 => h2?.textContent.trim(), h2);
+    text = await page.evaluate(h2 => h2?.textContent!.trim(), h2);
     expect(text).toBe('#Term dynamic content');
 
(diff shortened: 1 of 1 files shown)


JounQin:
But actually it can never be `null` at runtime, I'd prefer using `!` instead, or just keep it as-is.

https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#value

Marukome0743:
> But actually it can never be `null` at runtime, I'd prefer using `!` instead, or just keep it as-is.
> 
> https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#value

Thanks.
I'm getting TypeScript error of `textContent` is possibly null when don't check null.

スクリーンショット 2025-07-11 132428

So I have replaced `?` with `!`, non-null assertion at [fa0edfd](https://github.com/web-infra-dev/rspress/pull/2370/commits/fa0edfd056772fd8a4fe2cd7b2f5bd9e23099039).

JounQin:
I mean, why `tsconfig.json` should be added at the first place if everything runs well?

I don't quite get the necessary of this PR.

Marukome0743:
> I mean, why `tsconfig.json` should be added at the first place if everything runs well?
> 
> I don't quite get the necessary of this PR.

Thanks, I misunderstood.
I removed `tsconfig.json`.

JounQin:
> I don't quite get the necessary of this PR.

@Marukome0743 Did you get some real issues with current codes? When I wrote the tests, these were no errors reported on my editor.

Marukome0743:
> > I don't quite get the necessary of this PR.
> 
> @Marukome0743 Did you get some real issues with current codes? When I wrote the tests, these were no errors reported on my editor.

I got TypeScript warning which [I raised comment](https://github.com/web-infra-dev/rspress/pull/2370#issuecomment-3060479454).
It's just type checking, test is clean and no issue.

JounQin:
image

I'm also saying type checking, maybe that's because we're using different TypeScript versions on editors, the `textContent` getter shows `string` on my side.

JounQin:
image

OK, it's wrong on TypeScript 5.8.2 which rspress is using.

Marukome0743:
スクリーンショット 2025-07-11 180711

Yes, it's same with TypeScript 5.8.2.
Thank you for great explanation!
Eventually, I know what you say.

Anyway, it's great experience for me to know new TypeScript 5.9.0.

Issue #2290, opened 2025-06-17
[Feature]: Remove "_html" field in PageIndexInfo


SoonIter:
### What problem does this feature solve?

We already have ssg html artifact; if users want to use "_html", they should directly obtain the ssg output from the afterBuild hook.

_html is created by remark and rehype, it does not render React components, so it is missing information.

https://github.com/web-infra-dev/rspress/blob/b6cc0f06a3787485dabb062d603d512a2a781ce9/packages/plugin-rss/src/createFeed.ts#L28

### What does the proposed API look like?

https://github.com/web-infra-dev/rspress/blob/b6cc0f06a3787485dabb062d603d512a2a781ce9/packages/core/src/node/runtimeModule/siteData/extractPageData.ts#L112

PR #2371, opened 2025-07-11
refactor!: remove `builderPlugins` and migrate it to `builderConfig.plugins`

SoonIter:
## Summary

refactor!: replace builderPlugins with builderConfig.plugins

## Checklist

### DIFF
--- a/packages/core/src/node/logger/hint.ts
+++ b/packages/core/src/node/logger/hint.ts
@@ -47,11 +47,9 @@ export async function hintThemeBreakingChange(customThemeDir: string) {
 
 export function hintBuilderPluginsBreakingChange(config: UserConfig) {
   if (
-    // @ts-ignore config.builderPlugins is removed in V2
-    config.builderPlugins &&
-    // @ts-ignore
+    // config.builderPlugins is removed in V2
+    'builderPlugins' in config &&
     Array.isArray(config.builderPlugins) &&
-    // @ts-ignore
     config.builderPlugins.length > 0
   ) {
     logger.error(


SoonIter:
We tend to change `builderConfig` to `rsbuild`. But currently, we cannot do this, as it would mean that the rspress plugin for V1 will be completely incompatible with v2...

```ts
defineConfig({
   rsbuild: {
     plugins: []
   }
})
```

packages/document/docs/en/api/config/config-build.mdx
```diff
@@ -195,15 +203,6 @@ export default defineConfig({
 
 After enabling this config, Rspress will check the links in the document based on the conventional routing table. If there is an unreachable link, the build will throw an error and exit.
 
-### markdown.mdxRs
```
JounQin, line 198: `@rspress/mdx-rs` has not been removed.

SoonIter: We have actually removed it in loader, so this config is not work actually

but we are still using mdxRs in obtaining searchIndex.

We will also remove it in the future. For now, it will have a certain impact on perf, and I haven't had a good plan to remove it yet.

https://github.com/web-infra-dev/rspress/pull/2280

SoonIter: There are many pre-tasks, with a certain amount of work involved.

https://github.com/web-infra-dev/rspress/issues/2290

packages/core/src/node/logger/hint.ts
```diff
@@ -44,6 +45,35 @@ export async function hintThemeBreakingChange(customThemeDir: string) {
   }
 }
 
+export function hintBuilderPluginsBreakingChange(config: UserConfig) {
+  if (
+    // @ts-ignore config.builderPlugins is removed in V2
+    config.builderPlugins &&
```
JounQin:
```suggestion
    'builderPlugins' in config &&
```

I think this coule avoid `//@ts-ignore` as much as possible.

By the way, I'd always prefer `//ts-expect-error` wherever possible.

SoonIter:
https://github.com/web-infra-dev/rspress/discussions/1891#discussioncomment-13729235

Timeless0911 (approved):

PR #2381, opened 2025-07-14
fix: use SvgWrapper in LinkCard & NavScreenMenuGroup

jbroma:
## Summary

This PR fixes the issue were `LinkCard` would fail to render when replacing theme assets (as demonstrated here: https://v2.rspress.rs/guide/advanced/custom-theme#4-custom-icon). This is because it was not using `SvgWrapper` like every other theme component

I've also noticed the same issue with `NavScreenMenuGroup` and also fixed it within this PR

# Related issue

n/a

## Checklist


SoonIter:
thanks❤

PR #2434, opened 2025-08-02
fix(route): safe decode path with %

holyfata:
## Related Issue

https://github.com/web-infra-dev/rspress/issues/2372

## Checklist

- [X] Tests updated (or not required).
- [X] Documentation updated (or not required).

### DIFF
--- a/packages/shared/src/runtime-utils/utils.ts
+++ b/packages/shared/src/runtime-utils/utils.ts
@@ -217,8 +217,8 @@ export const parseUrl = (
 
 // decodeURIComponent will throw error if the url is not valid
 // this function will replace the invalid characters with %25
-export function safeDecodeURIComponent(url: string) {
-  return decodeURIComponent(url.replace(/%(?![0-9A-Fa-f]{2})/g, '%25'));
+export function safeDecodeURIComponent(uri: string) {
+  return decodeURIComponent(uri.replace(/%/g, '%25'));
 }
 
 export function normalizeHref(url?: string, cleanUrls = false) {
--- a/packages/shared/src/runtime-utils/utils.ts
+++ b/packages/shared/src/runtime-utils/utils.ts
@@ -215,6 +215,12 @@ export const parseUrl = (
   };
 };
 
+// decodeURIComponent will throw error if the url is not valid
+// this function will replace the invalid characters with %25
+export function safeDecodeURIComponent(url: string) {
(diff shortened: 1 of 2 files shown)


packages/shared/src/runtime-utils/utils.ts
```diff
@@ -215,6 +215,12 @@ export const parseUrl = (
   };
 };
 
+// decodeURIComponent will throw error if the url is not valid
+// this function will replace the invalid characters with %25
+export function safeDecodeURIComponent(uri: string) {
+  return decodeURIComponent(uri.replace(/%/g, '%25'));
```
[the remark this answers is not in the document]
heyq02: This will cause error.

SoonIter (dismissed):
thanks❤

packages/shared/src/runtime-utils/utils.ts
JounQin: Doesn't it change the meaning of original `%*`? For example if I have a url with `%20` which represents a space ` `, after this change, a url contains `%20` will be changed as `%2520`...

cc @SoonIter

JounQin (changes requested):

packages/shared/src/runtime-utils/utils.ts
heyq02: I think this restores the original meaning of the %. For decodeURI, % is an escape character, but in the case of a%b.md, it simply represents the % character itself. To make decodeURI recognize the original meaning of this character, we need to replace % with %25 before using decodeURI.

JounQin: It makes `%20` unavailable anymore.

heyq02: I was thinking too simply. I will try to filter all % encodings.

heyq02: Here is my final solution 
* Processing strategy:
  * 1. First attempt direct decoding, return result if successful
  * 2. If failed, check each %xx sequence individually:
    *    - Check if it's a valid hexadecimal format
    *    - Try to decode the specific %xx sequence
    *    - If decoding fails (e.g., invalid UTF-8 sequence), replace with %25xx
    *    - If format is incorrect, replace % with %25
  * 3. Finally decode the fixed string

SoonIter: sorry, this algorithm is too expensive for us to maintain...

Do other frameworks have this problem, how do they solve it, and are there any available packages already?

PR #2443, opened 2025-08-05
fix(plugin-llms): some md files sometimes do not emit

SoonIter:
## Summary

fix(plugin-llms): some md files sometimes do not emit

## Checklist


packages/plugin-llms/src/plugin.ts
```diff
@@ -151,17 +155,10 @@ const rsbuildPluginLlms = ({
             ).toString();
           } catch (e) {
             // normalizeMdFile might have some edge cases, fallback to no flatten and plain mdx
-            logger.debug(e);
+            logger.debug('normalizeMdFile failed', pageData.routePath, e);
             mdContent = content;
-            return;
```
SoonIter, line 156: this `return` is wrong

Issue #2419, opened 2025-07-25
[Bug]: duplicate id attributes


jl917:
### Version

```sh
-
```

### Details

We are currently facing an issue related to duplicate id attributes on our official website.

The project uses id="root" as the main container element. However, each page's content also assigns ids to HTML headings, and in some cases, one of those headings is also given the id="root".

This violates the HTML unique identifier rule, which requires all id values to be unique within a document. As a result, it can cause unexpected behavior such as anchor link malfunctions and incorrect behavior when using getElementById in JavaScript.

### Reproduce link

https://rspress.rs/api/config/config-basic

### Reproduce Steps

1. chrome console inject htmlHint 

[...]

PR #2450, opened 2025-08-06
fix(DOM): change DOM selector from root to __rspress_root (#2419)

jl917:
## Related Issue

#2419

## Checklist

### DIFF
--- a/e2e/fixtures/i18n/index.test.ts
+++ b/e2e/fixtures/i18n/index.test.ts
@@ -190,7 +190,7 @@ test.describe('i18n test', async () => {
     await page.waitForTimeout(100);
     await page.click('.rspress-nav-menu-group-content a');
     await page.waitForTimeout(100);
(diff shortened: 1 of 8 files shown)


SoonIter:
Sorry for the late reply, this is a breaking change, which affects users who use CSS #root selector to override Rspress default-theme styles.

I plan to combine it with the other theme style update release, otherwise, it may have some impact on users.

jl917:
> Sorry for the late reply, this is a breaking change, which affects users who use CSS #root selector to override Rspress default-theme styles.
> 
> I plan to combine it with the other theme style update release, otherwise, it may have some impact on users.

I understand. I expect this change could have a significant impact, so please feel free to roll it out gradually.

SoonIter:
thanks, done in https://github.com/web-infra-dev/rspress/issues/2533 and 2.0.0-rc.0

PR #2472, opened 2025-08-15
feat: enable rspack native watcher by default

GiveMe-A-Name:
## Summary

Rspack native watcher pass all rspress e2e tests. We can enable rspack native watcher in rspress by default. 

User can disable native watcher by setting RSPRESS_NATIVE_WATCHER as 'false' if we found any problems.

## Checklist

### DIFF
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -332,6 +332,7 @@ async function createInternalBuildConfig(
           .end();
 
         chain.experiments({
+          // Enable native watcher by default unless RSPRESS_NATIVE_WATCHER is set to 'false'
           ...chain.toConfig().experiments,
           nativeWatcher: process.env.RSPRESS_NATIVE_WATCHER !== 'false',
         });
--- a/packages/core/src/node/initRsbuild.ts
+++ b/packages/core/src/node/initRsbuild.ts
@@ -331,6 +331,11 @@ async function createInternalBuildConfig(
           })
           .end();
 
+        chain.experiments({
+          ...chain.toConfig().experiments,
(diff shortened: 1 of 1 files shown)


packages/core/src/node/initRsbuild.ts
```diff
@@ -318,6 +318,12 @@ async function createInternalBuildConfig(
           })
           .end();
 
+        chain.experiments({
+          // Enable native watcher by default unless RSPRESS_NATIVE_WATCHER is set to 'false'
+          ...chain.toConfig().experiments,
```
chenjiahan: `chain.toConfig()` is slow, we can use:

```suggestion
          ...chain.get('experiments'),
```

chenjiahan (approved):

packages/core/src/node/initRsbuild.ts
```diff
@@ -318,6 +318,12 @@ async function createInternalBuildConfig(
           })
           .end();
 
+        chain.experiments({
+          // Enable native watcher by default unless RSPRESS_NATIVE_WATCHER is set to 'false'
+          ...chain.get('experiments'),
+          nativeWatcher: process.env.RSPRESS_NATIVE_WATCHER !== 'false',
```
GiveMe-A-Name, line 337: Current rspress using `@rsbuild/core@1.4.15`, `@rsbuild/core@1.4.15` using `@rspack/core@1.4.11`
@rspack/core@1.4.11 has some bugs about nativeWatcher, will ignore FileSystem file change event.
rspack will solved it in 1.5.0.
So we need wait for rsbuild bump rspack version. then, rspress bump rsbuild version to solve it.

Timeless0911: Does Rsbuild 1.5.0-beta bring this fix?

PR #2512, opened 2025-08-26
feat(theme): Make it possible to mark elements to be excluded when copying

Karibash:
## Summary

To prevent the popup elements added by Twoslash from being copied along with the code, we’ve made it possible to mark elements that should be excluded during copying.

## Related Issue

#574

## Checklist


SoonIter (approved):

SoonIter:
This escape hatch is acceptable to me. Currently it is a private API, and we cannot guarantee that the class name will always remain unchanged (probably). :D We can also not include it in the document.

PR #2528, opened 2025-08-29
fix: command using object fully customize don't call normalize

zoolsher:
This pull request makes a small adjustment to the logic for handling package manager commands in the `PackageManagerTabs` component. The change ensures that when the `command` prop is already an object (with keys like `"yarn"`, `"pnpm"`, `"bun"`), the commands are not normalized, preserving any intentionally empty or custom values.

* Updated the `PackageManagerTabs` component to avoid normalizing commands when the `command` prop is an object with package manager keys, ensuring custom or empty commands are left unchanged.## Summary

## Checklist

### DIFF
--- a/packages/theme-default/src/components/PackageManagerTabs/index.tsx
+++ b/packages/theme-default/src/components/PackageManagerTabs/index.tsx
@@ -96,14 +96,14 @@ export function PackageManagerTabs({
     additionalTabs.forEach(tab => {
       commandInfo[tab.tool] = `${tab.tool} ${command}`;
     });
+    // Normalize yarn/pnpm/bun command
+    commandInfo.yarn && (commandInfo.yarn = normalizeCommand(commandInfo.yarn));
+    commandInfo.pnpm && (commandInfo.pnpm = normalizeCommand(commandInfo.pnpm));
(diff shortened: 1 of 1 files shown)


packages/theme-default/src/components/PackageManagerTabs/index.tsx
```diff
@@ -96,15 +96,15 @@ export function PackageManagerTabs({
     additionalTabs.forEach(tab => {
       commandInfo[tab.tool] = `${tab.tool} ${command}`;
     });
+    // Normalize yarn/pnpm/bun command
+    commandInfo.yarn && (commandInfo.yarn = normalizeCommand(commandInfo.yarn));
+    commandInfo.pnpm && (commandInfo.pnpm = normalizeCommand(commandInfo.pnpm));
+    commandInfo.bun && (commandInfo.bun = normalizeCommand(commandInfo.bun));
```
chenjiahan: It seems that we can remove `commandInfo.xxx &&` because they have already been defined above 😄

```suggestion
    commandInfo.yarn = normalizeCommand(commandInfo.yarn);
    commandInfo.pnpm = normalizeCommand(commandInfo.pnpm);
    commandInfo.bun = normalizeCommand(commandInfo.bun);
```

PR #2533, opened 2025-09-01
feat(theme)!: fully theme refactor

SoonIter:
## Summary

feat: theme refactor

## Related Issue

https://github.com/web-infra-dev/rspress/issues/2550

## Checklist

### DIFF
--- a/e2e/fixtures/i18n/index.test.ts
+++ b/e2e/fixtures/i18n/index.test.ts
@@ -200,15 +200,19 @@ test.describe('i18n test', async () => {
     expect(overviewContentZh ?? '').toContain('zh');
     expect(overviewContentZh ?? '').not.toContain('en');
 
-    const languageSwitcher = page.locator(
-      '.rp-nav__others .rp-nav-menu__item__container',
-    );
-    await languageSwitcher.first().click();
-    const englishOption = page.locator('.rp-hover-group__item a', {
+    const languageSwitcher = page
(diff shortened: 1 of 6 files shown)


packages/core/src/node/mdx/remarkPlugins/containerSyntax.ts
```diff
@@ -73,40 +77,21 @@ const createContainer = (
...
     type: 'containerDirective',
-    name: type,
+    name: 'Callout',
+    attributes: {
+      type: type,
+      title: title || getTypeName(type),
+    },
     data: {
-      hName: rootHName,
+      hName: 'Callout',
```
JounQin: Could it conflict with user defined global components? For example we have our own global `Callout` component in doom but for different meanings.

SoonIter: image

My answer is no I think.
You could have a try

JounQin: @SoonIter When there are global components, take `Callout` for example, `import { Callout } from '~/global-components'` will be injected on the top automatically, the compiled codes will be different:

image

image

So I believe they will be conflicted.

Could we change builtin component names starting with, for example, `$`: `$Callout`.

SoonIter: resolved by being reimplemented

PR #2536, opened 2025-09-01
fix(core): allow user can deploy to both `http://a.com/base/` and `http://a.com/base`

SoonIter:
## Summary

fix(core): "/base" should be rendered successfully when using `base`

When the base is set to ‘/’, `http://a.com/` and `http://a.com` point to the same page. (Browser default behavior)

image

However, when the base is set to ‘/base/’, `http://a.com/base/` and `http://a.com/base` actually represent different resources, but Rspress should allow access to `http://a.com/base` for compatibility, as users can deploy to /base. 

image

Rspress is compatible with both /base/ and /base

---

在 base 为 ‘/’ 时，`http://a.com/` 和  `http://a.com` 指向同一资源

但当 base 为 ‘/base/’ 时， `http://a.com/base/` 和 `http://a.com/base` 实际上代表着不同的资源，不过 Rspress 应该允许 `http://a.com/base` 的访问，以兼容用户可以部署到 /base 上

Rspress 兼容两种 /base/ 和 /base

## Some drawbacks

For relative URL links, they are very dangerous.

We use react-router-dom, which is a lenient routing matching mechanism.

For example, "/api" "/api/index" "/api/" all point to the same page, but different results will appear when they use relative path links, at this time, attention should be paid to

## Checklist

### DIFF
--- a/e2e/fixtures/with-base/index.test.ts
+++ b/e2e/fixtures/with-base/index.test.ts
@@ -43,4 +43,22 @@ test.describe('plugin test', async () => {
     const href = await page.evaluate(a => a?.getAttribute('href'), a);
     expect(href).toBe('/base/en/guide/install.html');
   });
+
+  test('Should render the homepage - "/base"', async ({ page }) => {
+    await page.goto(`http://localhost:${appPort}/base`, {
+      waitUntil: 'networkidle',
+    });
+    const docContent = await page.$('.rspress-doc');
+    const text = await docContent?.textContent();
+    expect(text?.includes('This is the index page')).toBeTruthy();
+  });
+
+  test('Should render the homepage - "/base/"', async ({ page }) => {
+    await page.goto(`http://localhost:${appPort}/base/`, {
+      waitUntil: 'networkidle',
+    });
+    const docContent = await page.$('.rspress-doc');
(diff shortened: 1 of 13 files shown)


e2e/fixtures/with-base/index.test.ts
```diff
@@ -43,4 +43,22 @@ test.describe('plugin test', async () => {
     const href = await page.evaluate(a => a?.getAttribute('href'), a);
     expect(href).toBe('/base/en/guide/install.html');
   });
+
+  test('Should render the homepage - "/base"', async ({ page }) => {
```
zoolsher, line 47: 有个小问题，我看前面是 build + preview ，build 会 SSG 所以就算没有 hydrate "This is the index page" 的文本内容也是有的

SoonIter: 考虑到了，包在 NoSSR 组件里，不会参与 SSG 过程，所以只有 hydration 结束会触发

Timeless0911 (approved):

e2e/fixtures/with-base/index.test.ts
zoolsher: Oh! I see!

PR #2545, opened 2025-09-02
fix: hmr not work 

SoonIter:
## Summary

fix: hmr error

## Checklist


packages/core/src/node/initRsbuild.ts
```diff
@@ -340,6 +340,7 @@ async function createInternalBuildConfig(
         if (chain.plugins.has(CHAIN_ID.PLUGIN.REACT_FAST_REFRESH)) {
           chain.plugin(CHAIN_ID.PLUGIN.REACT_FAST_REFRESH).tap(options => {
             options[0] ??= {};
+            options[0].test = [/\.([cm]js|[jt]sx?|flow)$/i, MDX_OR_MD_REGEXP];
```
chenjiahan, line 343: I didn't realize https://github.com/web-infra-dev/rsbuild/pull/5969 would break HMR in Rspress 😂 Maybe we could add some test cases for Rspress HMR

SoonIter: done by add e2e hmr case

Timeless0911 (approved):

PR #2549, opened 2025-09-03
feat(auto-nav-sidebar): support _meta.json and _nav.json hmr

SoonIter:
## Summary

feat(auto-nav-sidebar): support _meta.json and _nav.json hmr

img_v3_02pq_a42d8229-315d-49b0-a72a-b739a586cb6g

## Checklist


packages/core/src/node/auto-nav-sidebar/index.ts
```diff
@@ -35,39 +31,75 @@ export async function modifyConfigWithAutoNavSide(
       versions,
       config.root!,
       extensions,
+      metaFileSet,
+    );
+    return metaInfo;
+  }
+
+  if (hasLang) {
```
JounQin: What if the user wants `zh` lang without different locales support?

JounQin: What if the user wants `zh` lang without different locales support? Personally I think this check should be removed.

SoonIter: This is just the old copy logic; theoretically, this PR should only involve changes related to HMR, but I'm also happy to remove it

packages/core/src/node/auto-nav-sidebar/locales.test.ts
```diff
@@ -6,6 +6,10 @@ import {
 import { RouteService } from '../route/RouteService';
 import { processLocales } from './locales';
 
+function orderStringSet(input: Set<string>) {
```
Timeless0911, line 9: 1

packages/core/src/node/auto-nav-sidebar/walk.test.ts
```diff
@@ -21,10 +21,20 @@ RouteService.__instance__ = {
   getRoutePathParts: mockGetRoutePathParts,
 } as RouteService;
 
+function orderStringSet(input: Set<string>) {
```
Timeless0911, line 24: 2 duplicate

PR #2554, opened 2025-09-05
feat(twoslash): improve underline styling and adjust popup docs font-size

chenjiahan:
## Summary

Currently, twoslash's underline uses `currentColor` as its color, which appears a bit too dark. Therefore, I've converted it to a pseudo-element and added transparency, making it less visually prominent.

### Before

Screenshot 2025-09-05 at 21 50 14

### After

Screenshot 2025-09-05 at 22 02 06

## Checklist


SoonIter (approved):
LGTM

PR #2578, opened 2025-09-11
refactor: replace enhanced-resolve with rspack resolver

stormslowly:
## Summary
Using Rspack builtin Rspack-resolver for better performance.
https://rspack.rs/api/javascript-api/resolver#resolver-api

```ts
import { rspack } from '@rspack/core';

const { resolver } = rspack.experiments;

// Directly use Resolver API with default options
const { path: resolvedPath } = resolver.sync(contextPath, './index.js');
const { path: resolvedPath } = await resolver.async(contextPath, './index.js');

// Use ResolverFactory with custom options
const customResolver = new resolver.ResolverFactory(resolveOptions);
const result = customResolver.sync(contextPath, './request.js');
const result = await customResolver.async(contextPath, './request.js');
```

## Checklist

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -837,9 +837,6 @@ importers:
       chokidar:
         specifier: ^3.6.0
         version: 3.6.0
-      enhanced-resolve:
-        specifier: 5.18.3
-        version: 5.18.3
       github-slugger:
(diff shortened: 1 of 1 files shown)


SoonIter (approved):
thanks for your contribution ❤️️

PR #2580, opened 2025-09-11
feat(plugin-twoslash): Make the popup that is always displayed not flip its display position

Karibash:
## Summary

Since it appeared to flicker when scrolling, we made the popup that is always displayed not flip its display position.

Before

https://github.com/user-attachments/assets/1d14b467-b6b7-4e0d-b060-d9bbb98f9ace

After

https://github.com/user-attachments/assets/dc972236-ceb4-455d-b25e-cb33879f6b05

## Checklist


SoonIter:
LGTM, thanks~

Issue #2585, opened 2025-09-11
[Bug]: search with "con" breaks the Application


zoolsher:
### Version

```sh
System:
    OS: macOS 15.6.1
    CPU: (10) arm64 Apple M1 Pro
    Memory: 352.61 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 139.0.7258.155
    Edge: 140.0.3485.54
    Safari: 18.6
  npmPackages:
    @rspress/core: 2.0.0-beta.27 => 2.0.0-beta.27
    @rspress/plugin-algolia: 2.0.0-beta.27 => 2.0.0-beta.27
    @rspress/plugin-client-redirects: 2.0.0-beta.27 => 2.0.0-beta.27
    @rspress/plugin-llms: 2.0.0-beta.27 => 2.0.0-beta.27
    @rspress/plugin-rss: 2.0.0-beta.27 => 2.0.0-beta.27
    @rspress/plugin-sitemap: 2.0.0-beta.23 => 2.0.0-beta.23
```

### Details

- reproduce: search "con" word with lynx website
- error log:
```txt
lib-react.5daae3cf.js:2  TypeError: e[n].push is not a function
    at 90293.e8d27492.js:36:2117090
    at Array.reduce (<anonymous>)
    at L (90293.e8d27492.js:36:2117033)
    at M (90293.e8d27492.js:36:2117323)
    at 90293.e8d27492.js:36:2119109
    at Array.map (<anonymous>)
    at 90293.e8d27492.js:36:2119047
    at iE (90293.e8d27492.js:36:2119185)
    at ad (lib-react.5daae3cf.js:2:58533)
    at ud (lib-react.5daae3cf.js:2:72828)
```

### Reproduce link

https://lynxjs.org/

### Reproduce Steps

skip

zoolsher:
got the bug

<img width="1623" height="435" alt="Image" src="https://github.com/user-attachments/assets/25a34199-8755-4c63-af19-875693441b3e" />

PR #2586, opened 2025-09-11
fix: suggestions object key with constructor

zoolsher:
## Summary

This pull request refactors the way search suggestions are grouped and accessed in the `SearchPanel` component. The main change is switching from using a plain JavaScript object to a `Map` for grouping suggestions, which improves code clarity and consistency when handling dynamic keys.

**Refactor of suggestions grouping and iteration:**

* Changed the `normalizeSuggestions` function to return a `Map<string, DefaultMatchResultItem[]>` instead of a `Record<string, DefaultMatchResultItem[]>`, and updated the grouping logic to use `Map` methods.
* Updated the rendering logic to iterate over the keys and values of the `Map` using `Array.from(normalizedSuggestions.keys())` and `normalizedSuggestions.get(group)` instead of object property access.

## Related Issue

https://github.com/web-infra-dev/rspress/issues/2585

## Checklist

### DIFF
--- a/packages/theme-default/src/components/Search/SearchPanel.tsx
+++ b/packages/theme-default/src/components/Search/SearchPanel.tsx
@@ -361,7 +361,7 @@ export function SearchPanel({ focused, setFocused }: SearchPanelProps) {
         if (!groups.has(group)) {
           groups.set(group, []);
         }
-        groups.get(group)?.push(item);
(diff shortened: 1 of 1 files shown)


SoonIter:
Interesting, anothor option is `Object.create(null)`

Map is better than `Object.create(null)` I think. LGTM

PR #2589, opened 2025-09-11
feat(theme-default): Make the tab component usable in layouts other than DocLayout

Karibash:
## Summary

Before
スクリーンショット 2025-09-11 22 57 33

After
スクリーンショット 2025-09-11 22 59 03

## Checklist

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.2.2/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.2.3/schema.json",
   "assist": {
     "actions": {
       "source": {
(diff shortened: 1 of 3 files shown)


website/theme/index.tsx
```diff
@@ -30,16 +29,7 @@ function HomeLayout() {
...
-            containerElementClassName="language-bash"
-            codeButtonGroupProps={{
-              showCodeWrapButton: false,
-            }}
-          >
-            <Code className="language-bash" style={{ textAlign: 'center' }}>
-              npm create rspress@beta
-            </Code>
-          </PreWithCodeButtonGroup>
+          <PackageManagerTabs command="create rspress@beta" />
```
SoonIter, line 32: it was to show the fastest way to create, it doesn't need to be very complex I think 

ref: https://nextjs.org/

So could you please modify it to

```suggestion
          <PackageManagerTabs command={{ npm: "npm create rspress@beta" }} />
```

SoonIter: Oh, my fault, it may not be consistent with the original effect

PR #2623, opened 2025-09-25
feat(theme): add prop `dlx` to `PackageManagerTabs` for `npx` `pnpm dlx` etc

Nsttt:
## Summary

When using `<PackageManagerTabs>` with `exec={true}` it does not behave properly for `pnpm` and `bun`. This PR adds a new flag to distinguish between `exec` and `dlx`.

## Checklist

- [X] Tests updated (or not required).
- [X] Documentation updated (or not required).

### DIFF
--- a/packages/theme-default/src/components/PackageManagerTabs/index.tsx
+++ b/packages/theme-default/src/components/PackageManagerTabs/index.tsx
@@ -11,10 +11,16 @@ export type PackageManagerTabProps = (
   | {
       command: string;
       /**
-       * If true, the command will be interpreted as a shell command and prefixed with npx for npm,
-       * or the package manager binary for others.
+       * If true, uses local package execution (pnpm, bun, yarn, npm).
+       * For locally installed packages in node_modules.
        */
       exec?: boolean;
+      /**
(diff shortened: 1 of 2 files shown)


SoonIter:
we discussed in https://github.com/web-infra-dev/rspress/pull/2343 before

Nsttt:
> we discussed in #2343 before

Uh, you're actually right, I was a bit narrow minded.

So there's `pnpm/bun X` that works for local packages installed on closest `node_modules` use cases and there `pnpx`/`pnpm/bun dlx`  to execute remote packages. 

What do you think the best way to introduce this would would be ?

SoonIter:
> > we discussed in #2343 before
> 
> Uh, you're actually right, I was a bit narrow minded.
> 
> So there's `pnpm/bun X` that works for local packages installed on closest `node_modules` use cases and there `pnpx`/`pnpm/bun dlx` to execute remote packages.
> 
> What do you think the best way to introduce this would would be ?

Good point! Adding a new `dlx` prop sounds like a reasonable approach to handle this distinction.

PR #2722, opened 2025-11-05
chore(infra): add Rsdoctor bundle diff CI

yifancong:
## Summary
chore: rspress repo add diff CI

Demo: 
https://github.com/web-infra-dev/rspress/pull/2721

## Checklist

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -2,7 +2,9 @@ name: Diff
 
 on:
   pull_request:
-    types: [opened, synchronize, reopened, closed]
+    branches: [main]
+  push:
+    branches: [main]
 
 jobs:
   Build:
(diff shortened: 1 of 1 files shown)


.github/workflows/diff.yml
```diff
@@ -0,0 +1,48 @@
+name: Diff
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, closed]
```
Timeless0911: no limited pr branch? I think main is enough for diff CI, and types field is not need.

```
  pull_request:
    branches: [main]

  push:
    branches: [main]
```

BTW, is workflow_dispatch supported to run bundle diff CI thus we can diff manually with some commit ref.

yifancong: Done~ has been limited to the main branch

In addition, the proposal of workflow_dispatch is great, and I will support it after further verification.

SoonIter: > In addition, the proposal of workflow_dispatch is great, and I will support it after further verification.

+1

PR #2739, opened 2025-11-10
fix(theme): Outline font family was hardcoded

elliotcourant:
## Summary

The font family for the outline component title was hard coded as `Inter` but the base font family shipped with the docs doesn't include `Inter` instead including `Inter var` or others.

This could be changed to be `font-family: var(--rp-font-family-base);` but that seems redundant for this component where the only other places that overwrite the font family are for code blocks.

Current:

image

With Fix:

image

## Related Issue

N/A

## Checklist


SoonIter:
thanks for your contribution ❤

PR #2773, opened 2025-11-17
feat(plugin-twoslash): Make it possible to configure compilerOptions for TypeScript in twoslash

Karibash:
## Summary

Added `compilerOptions` option to `@rspress/plugin-twoslash` to allow customization of TypeScript compiler options used by Twoslash.
When displaying type information in code blocks with Twoslash, there are cases where you want to apply project-specific TypeScript compiler options. This feature allows you to flexibly customize the TypeScript environment used in code blocks within documentation.

## Usage Example

```ts
import { defineConfig } from '@rspress/core';
import { pluginTwoslash } from '@rspress/plugin-twoslash';
import { JsxEmit } from 'typescript';

export default defineConfig({
  plugins: [
    pluginTwoslash({
      compilerOptions: {
        jsx: JsxEmit.ReactJSX
      },
    }),
  ],
});
```

## Checklist

### DIFF
--- a/packages/plugin-twoslash/src/index.ts
+++ b/packages/plugin-twoslash/src/index.ts
@@ -9,8 +9,8 @@ import { fromMarkdown } from 'mdast-util-from-markdown';
 import { gfmFromMarkdown } from 'mdast-util-gfm';
 import { defaultHandlers, toHast } from 'mdast-util-to-hast';
 import picocolors from 'picocolors';
+import type { TwoslashOptions } from 'twoslash';
 import { removeTwoslashNotations } from 'twoslash';
-import type { CompilerOptions } from 'typescript';
 
 const staticPath = path.join(__dirname, '../static');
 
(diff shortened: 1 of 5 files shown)


packages/plugin-twoslash/src/index.ts
```diff
@@ -115,6 +116,10 @@ export interface PluginTwoslashOptions {
    * @default true
    */
   cache?: boolean;
+  /**
+   * TypeScript compiler options to use for Twoslash
+   */
+  compilerOptions?: CompilerOptions;
```
SoonIter: I prefer `twoslashOptions.compilerOptions`, it is possible that expose more properties from `twoslashOptions`.

Karibash: Would you prefer that we expose all of the `twoslashOptions`, or should we limit it to exposing only the `compilerOptions` for now?

SoonIter: Now my view is to expose the whole `twoslashOptions`, because of making it easier to explain to users and write documents.

Karibash: Fixed to expose twoslashOptions!
c6fe356

PR #2813, opened 2025-11-26
chore(theme): export `FallbackHeading` for reusing

JounQin:
## Summary

export `FallbackHeading` for reusing

We're customizing `Overview` component, `FallbackHeading` component is used for headings.

## Related Issue

N/A

## Checklist

### DIFF
--- a/packages/core/src/theme/components/FallbackHeading/index.tsx
+++ b/packages/core/src/theme/components/FallbackHeading/index.tsx
@@ -1,15 +1,17 @@
-import { renderInlineMarkdown } from '@theme';
+import { getCustomMDXComponent, renderInlineMarkdown } from '@theme';
 import { slug } from 'github-slugger';
-import { A } from '../DocContent/docComponents/a';
-import { H1, H2, H3, H4, H5, H6 } from '../DocContent/docComponents/title';
+import { useMemo } from 'react';
 
-const HEADING_MAP = {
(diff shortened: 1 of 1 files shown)


packages/core/src/theme/index.ts
```diff
@@ -17,7 +17,11 @@ export {
 export { getCustomMDXComponent } from './components/DocContent/docComponents/index';
 // layout
 export type { ShikiPreProps } from './components/DocContent/docComponents/pre';
-export { DocContent } from './components/DocContent/index';
+export {
```
SoonIter: Could we expose as a `@theme` components just like `Callout` `DocContent` etc instead of a sub folder of `DocContent`?

and `DocContent` import `FallbackHeading` through `@theme`

The reason is that we're ejecting the component source code using `rspress eject` now, so each component exported from `@theme` needs to be an independent unit.

JounQin: Done.

SoonIter: FallbackHeading is enough for users I think? Could we keep FallbackTitle as before?

SoonIter: I'll do this in this pull request

JounQin: Done.

SoonIter (approved):

PR #2832, opened 2025-12-03
chore(plugin-algolia): Add Russian translation for plugin-algolia

dragomano:
## Summary

Add Russian translation for plugin-algolia

Perhaps a new constant like RU_LOCALES should have been created? Or the existing one renamed? Please correct me if that's the case.


SoonIter:
Thanks ❤️️
Could you please disable the cspell of the translated text to fix the lint error?

PR #2841, opened 2025-12-03
feat(plugin-typedoc): Add setup callback for TypeDoc customization

### LINKED ISSUE #2642: [Feature]: Enhanced typedoc configuration

Currently it is impossible to configure typedoc properly since the settings aren't exposed. I would like to have a proper way to fine tune how things are generated.

You can decide your own. Or give me workarounds how I can do that.

Karibash:
## Summary

This PR adds a new `setup` option to the `@rspress/plugin-typedoc` plugin, allowing users to customize the TypeDoc Application before document generation.

### Example

```ts
pluginTypeDoc({
  entryPoints: ['./src/index.ts'],
  setup: (app) => {
    app.setOptions({
      categoryOrder: ['Core', 'Utils'],
    });
  },
});
```

## Related Issue

close #2642

## Checklist

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://biomejs.dev/schemas/2.2.3/schema.json",
+  "$schema": "https://biomejs.dev/schemas/2.3.8/schema.json",
   "assist": {
     "actions": {
       "source": {


packages/plugin-typedoc/src/index.ts
```diff
@@ -16,10 +16,15 @@ export interface PluginTypeDocOptions {
    * @default 'api'
    */
   outDir?: string;
+  /**
+   * A function to setup the TypeDoc Application.
+   * @param app
+   */
+  setup?: (app: Application) => Promise<void> | void;
```
SoonIter:
```suggestion
  typedoc?: (app: Application) => Promise<Application> | Promise<void> | void;
```

I think such an API would be more flexible if users want to replace the already generated typedoc instance in Rspress just like `tools.rspack`

- https://rsbuild.rs/zh/config/tools/rspack#function-%E7%B1%BB%E5%9E%8B

```ts
import { Application } from 'typedoc';

pluginTypeDoc({
  typedoc: async () => {
    return Application.bootstrapWithPlugins({
        name: config.title,
        entryPoints,
        disableSources: true,
        router: 'kind',
        readme: 'none',
        githubPages: false,
        requiredToBeDocumented: ['Class', 'Function', 'Interface'],
        // @ts-expect-error - Typedoc does not export a type for this options
        plugin: [loadPluginMarkdown],
        entryFileName: 'index',
        hidePageHeader: true,
        hideBreadcrumbs: true,
        pageTitleTemplates: {
          module: '{kind}: {name}', // e.g. "Module: MyModule"
        },
      });
  }
})
```

Karibash: @SoonIter 
Thanks! I addressed this in 32bbd73.

I kept the property name as setup instead of typedoc.
Since this property belongs to pluginTypeDoc, I think setup is more intuitive in this context.

Let me know if you'd prefer a different naming approach.

SoonIter:
thanks ❤️️

PR #2860, opened 2025-12-08
feat(theme)!: custom theme should use `@rspress/core/theme-original`

SoonIter:
### Before

> In Rspress V1 document, `@theme` is the alias for `rspress/theme`.

but `@theme` is an alias which points to `theme/index.tsx` and falls back to `rspress/theme`

 `rspress/theme` or `@rspress/core/theme` points to Rspress's `theme-default`, they are actually different things.
 

### After

We standardize this behavior, which is the same as docusaurus.

1. In `docs` directory, `@theme` or `@rspress/core/theme` should be used, which points to `theme/index.tsx` and falls back to `theme-default`

2. In `theme` directory ,  `@theme-original` or `@rspress/core/theme-original`  should be used, which always points to Rspress's `theme-default`

## For `@theme-assets`

- https://x.com/_jbroma/status/1945517952708567477

### Before

We have a separate entry for exporting @theme-assets

```tsx
import { SvgWrapper } from '@theme'
import SearchSvg from '@theme-assets/Search';

<SvgWrapper icon={SearchSvg} />
```

### After

`@theme-assets` has been removed. Icons are exported from `@theme` and prefixed with `Icon`, for example: `IconSearch`

```tsx
import { SvgWrapper, IconSearch } from '@theme';

<SvgWrapper icon={IconSearch} />
                   ^? type Icon = React.FC<React.SVGProps<SVGSVGElement>> | string;
```

## Checklist

### DIFF
--- a/e2e/fixtures/custom-icon/theme/env.d.ts
+++ b/e2e/fixtures/custom-icon/theme/env.d.ts
@@ -1,0 +1,4 @@
+declare module '*.svg' { 
+  const icon: string;
+  export default icon;
+}
--- a/e2e/fixtures/custom-icon/theme/index.tsx
+++ b/e2e/fixtures/custom-icon/theme/index.tsx
@@ -1,0 +1,4 @@
+import IconSearch from './assets/search.svg';
+
+export { IconSearch };
+export * from '@rspress/core/theme-original';


SoonIter:
image

https://docusaurus.io/docs/swizzling#wrapping

PR #2866, opened 2025-12-08
fix(core): Fix an issue where llms.txt was not generated when locales was not set

Karibash:
## Summary

If locales is not set in the config, getLangs returns an empty array.
As a result, llms.txt was not being generated.

## Checklist


SoonIter (approved):

SoonIter:
thanks❤️️

Issue #1552, opened 2024-11-05
[Feature]: Use React Routers 6.27 `viewTransition` prop


chenjiahan:
### What problem does this feature solve?

Since React Router 6.27.0, using View Transition API requires only [adding viewTransition prop to Link component](https://reactrouter.com/en/6.27.0/components/link#viewtransition):

```js
<Link to={to} viewTransition>
  Click me
</Link>
```

It can replace the current `useViewTransition` hook in Rspress, and this should fix the current runtime error logs when using `themeConfig.enableContentAnimation`:

![Image](https://github.com/user-attachments/assets/856d97cb-231d-47f9-a595-7f97f90cc023)

### What does the proposed API look like?

- Using the `viewTransition` prop
- The React Router document also notes that `viewTransition` only works when using a data router, see [Picking a Router](https://reactrouter.com/en/6.27.0/routers/picking-a-router).

chenjiahan:
Update: Consider https://react.dev/reference/react/ViewTransition

PR #2898, opened 2025-12-17
chore: remove unused view transition code

SoonIter:
## Summary

Currently, `enableContentAnimation` is not implemented, so this configuration is currently invalid.

First, delete the code that is not currently used, and it can be added back later.

## Checklist


SoonIter:
some new APIs can be considered

https://github.com/web-infra-dev/rspress/issues/1552

naico-wang:
image
Always get this in my dev environment, maybe some issue caused by useViewTransition hook

PR #2912, opened 2025-12-21
chore(plugin-llms): Add Russian translation for plugin-llms

dragomano:


dragomano:
Is there a way to globally disable the linter for the Russian language or fine-tune it so that I don't have to add comments to every file I modify?

PR #2923, opened 2025-12-24
fix(core): `Iterator#map` is only available on Node 22+

JounQin:
## Summary

`Iterator#map` is only available on Node 22+ which is incompatible with rspress's baseline [`>=18.0.0`](https://github.com/web-infra-dev/rspress/blob/ddb52660e348dc09d0867a1e07b4f418628a066d/packages/core/package.json#L146C14-L146C22)

## Related Issue

N/A

## Checklist


SoonIter:
@Timeless0911 How can I ensure that the output compiled by Rslib corresponds to my engines field?

PR #2926, opened 2025-12-24
feat(auto-nav-sidebar): support download links in nav

JounQin:
## Summary

support download links in nav

## Related Issue

kind of related to #2822, I think changing `nav` at runtime would be a better solution for our case, but download link is not supported yet

## Checklist

### DIFF
--- a/packages/core/src/theme/components/HoverGroup/index.tsx
+++ b/packages/core/src/theme/components/HoverGroup/index.tsx
@@ -69,7 +69,8 @@ function HoverGroupItem({
   }
 
   if ('link' in item) {
-    const { text, link, lang, rel, download } = item;
+    const { text, link, lang, rel } = item;
+    const download = 'download' in item ? item.download : undefined;
     const isActiveItem = activeMatcher
       ? activeMatcher(item)
       : matchNavbar(item, pathname);


packages/shared/src/types/theme/nav.ts
```diff
@@ -10,6 +10,7 @@ export type NavItem =
 export type NavItemWithLink = {
   text: string;
   link: string;
+  download?: boolean;
   tag?: string;
   activeMatch?: string;
```
JounQin, line 15: Hmm... Just found it's impossible to match different external links here(in our case, versions are external, `/v1`, `/v2`, `/v3`, etc), maybe functional `activeMatch` could be supported?

SoonIter:
...emmm, It is relatively difficult to maintain,  this feature should also be added `_nav.json` 

how about this workaround

```ts
// theme/index.tsx
import { Link as BasicLink } from '@rspress/theme/original'

const Link = (props) => {
  if(props.href === "mylink") {
    return <a></a>
  }
  return <BasicLink {...props} />
}

export { Link }
export * from '@rspress/core/theme-original';
```

JounQin:
> this feature should also be added `_nav.json`

I didn't read the source codes carefully yet, but doesn't it automatically supported?

> how about this workaround

We fetch data at runtime and change nav dynamically, and that also doesn't work with current `activeMatch` which could not match external links like `/external/v1` as I pointed at https://github.com/web-infra-dev/rspress/pull/2926#discussion_r2645354295.

JounQin:
Did real hack way to implement without changes. https://github.com/alauda/doom/pull/231

JounQin:
@SoonIter I hope this feature can be reconsidered, we tried your proposed workaround, but found it's hard to distinguish which links should be `download` actually, in our case, the nav download links are dynamic generated across multiple sites. And `download` affects browser behavior just like #3132, so I personally think it's important to be supported.

SoonIter (dismissed):

PR #2983, opened 2026-01-09
fix(nav): Don't include empty sidebar directories

elliotcourant:
## Summary

This fixes an issue where if a directory does not contain any valid files that would produce a route in the sidebar then that directory is not included in the sidebar automatically. But it should still be included in the sidebar if the route is named outright in the `_meta.json`. This way you don't end up with a sidebar item for an empty directory that may be accidental or may not be intended for text content.

An example of this is when you are co-locating image assets with the markdown documents they go with. Creating the following structure:

- `docs/foo.md`
- `docs/assets/foo_screenshot_1.png`

Before this commit, there would be an `assets` sidebar item that would be a collapsible directory with no items. With this change, the `assets` item would now be omitted by default unless it contains a valid index file or other valid text content or is included in the `_meta.json` deliberately.

## Related Issue

There is no related issue, however if you would like I can create one with a reproduction link. However if you revert my change in `packages/core/src/node/auto-nav-sidebar/normalize.ts` and then run `pnpm test:unit` you will see the test fail with an empty assets directory item.

## Checklist

  - Covered by existing tests, to see this run these commands on this branch:
    ```shell
    # Revert code fix
    git checkout main packages/core/src/node/auto-nav-sidebar/normalize.ts
    pnpm test:unit # Tests will fail
    # Re-add code fix
    git checkout HEAD packages/core/src/node/auto-nav-sidebar/normalize.ts
    pnpm test:unit # Tests will pass
    ```

### DIFF
--- a/packages/core/src/node/auto-nav-sidebar/normalize.ts
+++ b/packages/core/src/node/auto-nav-sidebar/normalize.ts
@@ -132,9 +132,10 @@ async function metaItemToSidebarItem(
       mdFileSet,
       false,
     );
-    // If a directory contains no valid markdown files or files are being
-    // excluded from a directory such that there are no valid items to have in
-    // the sidebar for a directory, then don't include that directory at all.
+    // If an automatically discovered directory (i.e. one not explicitly
+    // configured in `_meta.json`) contains no valid markdown files, or files
+    // are excluded such that there are no valid items for the sidebar, then
+    // don't include that directory at all.
     if (dir.items.length === 0) {
       return [];
     }


elliotcourant:
This is actually wrong, if you do include the `assets` dir in the `_meta.json` file this still excludes it. I need to somehow move this check for directories up? Or maybe excluding empty directories from the auto side nav is still desirable behavior but then I need to update the commit notes as to not confuse anybody in the future.

Let me know if this is sill desirable behavior or if I should move my check upward in the code such that it still does allow an empty directory to be "force added" via the `_meta.json` file.

elliotcourant:
Closing for now, there's more here for me to understand before this is actually ready

PR #3000, opened 2026-01-13
fix(config/types): defineConfig should support asyncFn

SoonIter:
## Summary

fix(config/types): defineConfig should support asyncFn

## Checklist


Timeless0911 (approved):

SoonIter:
Twoslash introduces JSDocs, which causes the bundle size to grow.

Issue #3068, opened 2026-01-27
[Bug]: Toggle button still visible when Dark Mode is disabled


maccuaa:
### Version

```sh
System:
    OS: macOS 15.7.3
    CPU: (12) arm64 Apple M4 Pro
    Memory: 110.72 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 144.0.7559.97
    Edge: 144.0.3719.92
    Safari: 26.2
  npmPackages:
    @rspress/core: 2.0.0-rc.7 => 2.0.0-rc.7
    @rspress/plugin-rss: 2.0.0-rc.7 => 2.0.0-rc.7
    @rspress/plugin-sitemap: 2.0.0-rc.7 => 2.0.0-rc.7
```

### Details

When I disable Dark Mode in the theme config, the toggle button is still displayed in the Nav Bar.

**rspress.config.ts**
```ts
themeConfig: {
    darkMode: false,
}
```

Looks like this only affects the Nav component - [Nav/index.tsx:L51](https://github.com/web-infra-dev/rspress/blob/28a333a58f81d65482e6fa1d7f77e76b4e252ef5/packages/core/src/theme/components/Nav/index.tsx#L51)

The other Nav components have a check to conditionally render the SwitchAppearance component:

```ts
const hasAppearanceSwitch = site.themeConfig.darkMode !== false;
```

**NavScreenMenuOthers**

[NavScreenMenuOthers.tsx#L14-L20](https://github.com/web-infra-dev/rspress/blob/28a333a58f81d65482e6fa1d7f77e76b4e252ef5/packages/core/src/theme/components/NavScreen/NavScreenMenuOthers.tsx#L14-L20)

**NavScreenAppearance**

[NavScreenAppearance.tsx#L11-L20](https://github.com/web-infra-dev/rspress/blob/28a333a58f81d65482e6fa1d7f77e76b4e252ef5/packages/core/src/theme/components/NavScreen/NavScreenAppearance.tsx#L11-L20)

### Reproduce link

N/a

### Reproduce Steps

1. disable Dark Mode in theme config
2. start dev server

Expected Result:
Theme toggle is not displayed

Actual Result:
Theme toggle is displayed

SoonIter:
closed by https://github.com/web-infra-dev/rspress/pull/3069 and released in V2.0.0

have a nice day ❤, thanks~

PR #3069, opened 2026-01-27
fix(Nav): conditionally render appearance switch based on darkMode setting (#3068)

maccuaa:
## Summary

Don't render the SwitchAppearance component in the Nav bar when Dark Mode is off.

## Related Issue

#3068 

## Checklist

- [x] Documentation updated (**not required**).


SoonIter (approved):
thanks ❤

PR #3079, opened 2026-01-30
feat: consider mf async entry scene

2heal1:
## Summary

Use MF will change the entry as Promise<PreviousEntry>, so need to add await

## Checklist


SoonIter:
If MF is stable, you can add an e2e test to avoid frequent searches for similar issues.

PR #3115, opened 2026-02-09
feat(SEO/custom-head): Make `og:description` set automatically

### LINKED ISSUE #3114: [Feature]: Automatically set `og:description` from page description

Currently, the page `title` is automatically used for `og:title`, which is very convenient.

However, the page `description` is **not** reflected in `og:description` by default.

While `rsbuild-plugin-open-graph` can be used to set `og:description`, it applies the **same static value to all pages**, which makes it unsuitable for pages that have per-page descriptions.

This creates an inconsistency:
- `og:title` is automatically derived from the page title
- `og:description` must be manually configured and cannot easily vary per page

Ideally, `og:description` should behave the same way as `og:title`:

- If a page has a `description`, use that value for `og:description`
- This should work on a per-page basis, without requiring a global static configuration

This would make Open Graph metadata more consistent and reduce the need for additional plugins or manual configuration.

Karibash:
## Summary

Like og:title, og:description is now set automatically.

## Related Issue

close #3114

## Checklist

### DIFF
--- a/website/docs/en/guide/advanced/custom-head.mdx
+++ b/website/docs/en/guide/advanced/custom-head.mdx
@@ -4,12 +4,13 @@ Adding head tags (such as meta tags) to web pages is crucial for SEO optimizatio
 
 Currently, Rspress automatically injects the following head tags:
 
-| Tag                                                     | Description                                                                         |
-| ------------------------------------------------------- | ----------------------------------------------------------------------------------- |
-| `<meta name="generator" content="Rspress v${version}">` | Identifies the site generator and version                                           |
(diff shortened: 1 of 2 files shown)


SoonIter (changes requested):
thanks❤️️, could you update the document here?

- https://rspress.rs/guide/advanced/custom-head

Karibash:
@SoonIter Thanks! I’ve updated the document at the link you mentioned. f6b05ab

SoonIter (approved):

SoonIter:
thanks, very useful changes

PR #3117, opened 2026-02-09
fix(theme): avoid selection gaps at line ends in code block

KazariEX:
## Summary

| before | after |
| --- | --- |
| image | image |

## Related Issue

*no response*

## Checklist

### DIFF
--- a/packages/core/src/theme/components/CodeBlock/index.scss
+++ b/packages/core/src/theme/components/CodeBlock/index.scss
@@ -50,9 +50,8 @@
     // #endregion text
 
     position: relative;
-    width: fit-content;
     margin: 0;
-    padding: 1rem 1.25rem;
+    padding: 1rem 0;
     background: transparent;
     z-index: 1;
   }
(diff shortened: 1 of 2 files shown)


SoonIter:
pending by two issues, it seems a breaking change for all the transformers

1. twoslash

image

2. transformerNotationDiff

image


## Before

```css
.highlighted {
   background-color: rgba(59, 130, 246, 0.1);
    box-shadow: inset 2px 0 0 0 rgb(59, 130, 246);
 }
```

## after

This makes the definition of the transformer's CSS more complicated, and almost **all related transformers** will become more complicated than before. I'm not sure if it's worth it, but it's the behavior in vitepress by the Shiki....

```css
.highlighted {
    width: calc(100% + 2.5rem);
    margin: 0 -1.25rem;
    padding: 0 1.25rem;
	background-color: rgba(59, 130, 246, 0.1);
    box-shadow: inset 2px 0 0 0 rgb(59, 130, 246);
 }
```

image

KazariEX:
We can add this selector to elements that occupy an entire row:

```css
:where(.highlighted, ...) {
    width: calc(100% + 2.5rem);
    margin: 0 -1.25rem;
    padding: 0 1.25rem;
}
```

PR #3006, opened 2026-01-14
feat(core): auto-extract description from first contentful line in extractPageData

SoonIter:
## Summary

Add automatic description extraction to `extractPageData` when `frontmatter.description` is not provided. The description is extracted from the first paragraph before the first h2 heading in the markdown content.

> https://docusaurus.io/docs/markdown-features/head-metadata#markdown-page-description

image

## Checklist

- [x] Tests updated (or not required).
- [ ] Documentation updated (or not required).

---

### Changes Made

1. **Added `description` field to `PageIndexInfo` type** (`packages/shared/src/types/index.ts`)
   - New optional `description?: string` field added to the interface

2. **Implemented description extraction in `extractPageData.ts`** (`packages/core/src/node/route/extractPageData.ts`)
   - Added `extractTextFromNode()` function: recursively extracts plain text content from AST nodes
   - Added `extractDescription()` function: extracts the first paragraph's text content before any h2 heading
   - The `description` field is set at the top level of `PageIndexInfo`, while `frontmatter.description` preserves the original frontmatter value

3. **Added test cases** (`packages/core/src/node/route/extractPageData.test.ts`)
   - Tests for automatic description extraction when `frontmatter.description` is not set
   - Tests to verify `frontmatter.description` takes priority when provided
   - Tests to ensure `frontmatter.description` remains unchanged (preserves original value)

### How It Works

- When processing a markdown file, if `frontmatter.description` exists, use it as the `description` field value
- If `frontmatter.description` is not set, extract the first paragraph before the first h2 heading as the description
- The `frontmatter.description` always preserves the original frontmatter value (not modified by extraction)

This PR was written using [Vibe Kanban](https://vibekanban.com)

---

PR #3134, opened 2026-02-11
feat(plugin-typedoc): Add description to frontmatter in generated docs

Karibash:
## Summary

This PR enhances the plugin-typedoc to automatically add description metadata to the frontmatter of generated API documentation pages.

## Checklist

### DIFF
--- a/pnpm-workspace.yaml
+++ b/pnpm-workspace.yaml
@@ -7,7 +7,8 @@ packages:
   - '!**/tests/**'
   - '!packages/create-rspress/template-basic'
 
-hoistPattern: []
+hoistPattern:
+  - typedoc
 
 linkWorkspacePackages: true
 
--- a/e2e/fixtures/plugin-typedoc/index.test.ts
+++ b/e2e/fixtures/plugin-typedoc/index.test.ts
@@ -6,6 +6,7 @@ import { getPort, killProcess, runDevCommand } from '../../utils/runCommands';
 test.describe('plugin-typedoc single entry', async () => {
   let appPort: number;
   let app: Awaited<ReturnType<typeof runDevCommand>>;
+
   test.beforeAll(async () => {
     const appDir = path.join(__dirname, 'single');
     appPort = await getPort();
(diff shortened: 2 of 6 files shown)


pnpm-workspace.yaml
```diff
@@ -1,21 +1,19 @@
...
-  - core-js
-  - esbuild
-  - nx
-  - sharp
+  - 'core-js'
+  - 'esbuild'
+  - 'nx'
 
+hoistPattern:
+  - 'typedoc'
```
Karibash: Because [typedoc-plugin-frontmatter does not list typedoc as a dependency](https://github.com/typedoc2md/typedoc-plugin-markdown/blob/fe732a57a2ea90b937d0b0b67e1ef53531c7437b/packages/typedoc-plugin-frontmatter/package.json#L33-L35), an error was occurring during dependency resolution. This configuration is intended to hoist typedoc as a workaround.

SoonIter:
- https://github.com/web-infra-dev/rspress/pull/3006

It seems to overlap with this feature. Can using this feature meet the requirement?

> https://rspress.rs/api/config/config-build#markdownextractdescription

Adding typedoc plugin will increase some maintenance costs and the number of dependencies, if using extractDescription can meet the requirements, I think it's better to use extractDescription

Karibash:
@SoonIter 
Thank you for pointing this out. I didn’t know that such a great feature already existed.

However, I’m not sure if extractDescription works well for the documentation generated by TypeDoc. For example, when TypeDoc generates documentation like in the attached screenshot, the extracted description ends up looking like the one shown in the second screenshot. In this case, it includes some information that should not be part of the description.
スクリーンショット 2026-02-13 22 01 36
スクリーンショット 2026-02-13 22 01 45

On the other hand, with the implementation in this PR, the generated description looks like the one in the third screenshot, which better matches the intended content.
スクリーンショット 2026-02-13 22 02 07

So, while I agree that using the built-in feature would be preferable if it meets the requirements, I think there may still be some limitations when it comes to TypeDoc-generated content. Please let me know what you think.

Karibash:
I reviewed the Markdown generated by TypeDoc, and it seems that this issue might be resolved by ignoring the quoted sections.

Would it be better to close this PR and proceed with a fix that makes extractDescription ignore quoted content?

SoonIter:
> I reviewed the Markdown generated by TypeDoc, and it seems that this issue might be resolved by ignoring the quoted sections.
> 
> Would it be better to close this PR and proceed with a fix that makes extractDescription ignore quoted content?

Yeah, I think optimizing the extractDescription logic is the better path forward. Let's go with a targeted fix instead.

PR #3160, opened 2026-02-26
feat(theme): add ScrollRestoration to avoid scroll flash

SoonIter:
## Summary

Due to SSG + hydration, there is a scroll flash on the first screen. We used a common community solution to implement Rspress's own ScrollRestoration

https://reactrouter.com/api/components/ScrollRestoration

- Add `ScrollRestoration` component that saves and restores scroll positions on browser back/forward navigation
- Renders an inline `<script>` tag for pre-hydration scroll restoration (avoids flash)
- Uses React Router's `location.key` as storage key, persists to `sessionStorage` via `pagehide`
- Skip hash-based anchor scrolling on POP navigation to avoid overriding restored position

Inspired by React Router's `<ScrollRestoration>` component.

## Test plan

- [ ] Navigate to a long doc page, scroll down, click a link, press browser back — scroll position should be restored
- [ ] Navigate to a new page — should scroll to top
- [ ] Click a hash/anchor link — should scroll to the anchor element
- [ ] Reload the page — scroll position should be restored from sessionStorage

### DIFF
--- a/packages/core/src/theme/logic/ScrollRestoration.tsx
+++ b/packages/core/src/theme/logic/ScrollRestoration.tsx
@@ -5,7 +5,10 @@ const STORAGE_KEY = 'rspress-scroll-positions';
 const MAX_SCROLL_ENTRIES = 100;
 
 // Module-level state for scroll positions
-let savedScrollPositions: Record<string, number> = {};
+const savedScrollPositions: Record<string, number> =
+  typeof window === 'undefined'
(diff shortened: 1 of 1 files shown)


SoonIter:
- https://github.com/ipatate/react-router-scroll-memory
- https://api.reactrouter.com/v7/functions/react-router.ScrollRestoration.html

PR #3162, opened 2026-02-27
perf(theme-default): defer main content rendering on navigation

Huxpro:
## Summary

When pages have heavy MDX content, rendering the sidebar and content together in one React commit causes jank during client-side navigation. This PR uses React 18's `useTransition` to split rendering into two commits:

1. **Commit 1**: Sidebar renders immediately, giving instant navigation feedback
2. **Commit 2**: Doc content renders as a non-urgent transition, allowing React to time-slice and keep the UI responsive

SSR hydration is unaffected — content renders immediately on initial load since the deferral only activates on pathname changes.

## Related Issue

N/A

## Checklist

### DIFF
--- a/packages/core/src/theme/layout/DocLayout/index.tsx
+++ b/packages/core/src/theme/layout/DocLayout/index.tsx
@@ -1,4 +1,4 @@
-import { useFrontmatter } from '@rspress/core/runtime';
+import { useFrontmatter, useLocation } from '@rspress/core/runtime';
 import {
   DocContent,
   DocFooter,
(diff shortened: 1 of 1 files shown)


SoonIter:
It seems we already have enabled React useTransition in Sidebar. I followed the React official document. I'm not sure if it is related to this PR, 

- https://github.com/web-infra-dev/rspress/pull/2778
- https://react.dev/reference/react/useTransition#usage

> Commit 1: Sidebar renders immediately, giving instant navigation feedback

We already have the UI display related to isPending, and it is executed immediately as this GIF sees

20260227193723_rec_-convert



and using two commits causes a flash in my local test

20260227200418_rec_-convert

PR #3163, opened 2026-02-27
feat(theme/llms): add configurable placement option for LLM UI buttons

Huxpro:
## Summary

Add a `placement` option to `llmsUI` config that controls where LLM UI components (copy markdown, open in chat) are displayed:

- `'title'` (default): buttons below the H1 title, preserving the original behavior
- `'outline'`: separate rows in the right-side outline panel, similar to "Scroll to top"

New outline components (`LlmsCopyRow`, `LlmsOpenRow`) are added with styling that matches existing outline bottom items. Documentation updated across all relevant pages (en/zh).

## Checklist


Huxpro:
image

image

SoonIter:
Thanks for the effort here, @Huxpro

The LLM UI buttons are intentionally placed near the title (and below the title on mobile / after the breadcrumb), following a common convention across documentation frameworks. Users have likely already developed muscle memory around this location, so adding an alternative placement may introduce confusion rather than convenience.


- fumadocs(same as Rspress)

https://www.fumadocs.dev/docs/guides/rss

image

- bun,  claude (minitlify)

https://bun.com/docs/installation

https://platform.claude.com/docs/en/intro

image
image

- nuxt ui

https://ui.nuxt.com/docs/getting-started

image


- nextjs 

https://nextjs.org/docs/app/getting-started/project-structure

image


After looking into this, I think we'll keep the current behavior as-is for now. If you have a specific use case where the current placement falls short, it is more appropriate to configure through custom themes I think 🤔

Huxpro:
@SoonIter 

https://elements.ai-sdk.dev/ put it on outline and I want it on lynxjs.org:

image

I also made it configurable and default to "title" so it doesn't change the original behavior, right? It doesn't hurt to have more options

SoonIter:
> @Huxpro 
> [elements.ai-sdk.dev](https://elements.ai-sdk.dev/) put it on outline and I want it on lynxjs.org:
> I also made it configurable and default to "title" so it doesn't change the original behavior, right? It doesn't hurt to have more options

Get ❤, thanks

SoonIter:
In addition to this PR, a fix will also be made. Rspress will optimize scrolling for overly long outlines like elements.al.


from (height: auto)

image

to (scrollable in one small container with fixed height)

image



I'll let agent to do this task

SoonIter (approved):

Issue #1527, opened 2024-10-25
[Bug]: Anchor `#` for titles in <Steps> overlaps with the sequence number


i-richardwang:
### Version

System:
OS: macOS 15.0.1
CPU: (10) arm64 Apple M1 Pro
Memory: 213.81 MB / 16.00 GB
Shell: 3.2.57 - /bin/bash
Browsers:
Chrome: 130.0.6723.58
npmPackages:
rspress: ^1.34.1 => 1.34.1

### Details

Anchor `#` for titles in <Steps> overlaps with the sequence number

![Image](https://github.com/user-attachments/assets/6ebd7672-4943-437e-ac9e-d47f8bcad235)

### Reproduce link

None

### Reproduce Steps

```
import { Steps } from '@theme';

<Steps>

### Title1

1. node1
2. node2
3. node3

</Steps>
```

Timeless0911:
Seems can not be reproduced in website https://rspress.dev/guide/default-theme/components#steps

JounQin:
> Seems can not be reproduced in website [rspress.dev/guide/default-theme/components#steps](https://rspress.dev/guide/default-theme/components#steps)

<img width="1204" alt="Image" src="https://github.com/user-attachments/assets/b4bd87eb-ca3c-44e9-b29c-104c143247ca" />

This can be reproduced very easily.

Timeless0911:
Rspress set `markdown.mdxRs: true` by default and mdx-rs will not add header anchor for headings in components which may be not aligned with mdxjs.

SoonIter:
Closing this issue as it has been inactive for over a year. If this is still relevant, please feel free to reopen it or create a new issue with updated context. Thank you for your contribution!

SoonIter:
Now this behavior is retained in V2, and I think it is acceptable.

1. It can be entered into the TOC.

2. Display # for jump.

- https://rspress.rs/ui/components/steps

<img width="1572" height="931" alt="Image" src="https://github.com/user-attachments/assets/82dd7acf-7de2-47df-b10a-9d6796fc8072" />

PR #3276, opened 2026-04-04
fix(theme/Steps): # for titles overlaps with the sequence number

elliotcourant:
## Summary

When inside a stepper component, the anchor link should render at the end of the header instead of at the beginning in order to avoid strange z-fighting.

### Before

https://github.com/user-attachments/assets/7ad3abfc-d1cc-4e8e-8f82-7dcc2102899b

### After

https://github.com/user-attachments/assets/8d437edd-df1d-4be7-9257-47d9166673e4

## Related Issue

N/A

## Checklist

### DIFF
--- a/packages/core/src/theme/components/Steps/index.scss
+++ b/packages/core/src/theme/components/Steps/index.scss
@@ -1,51 +1,54 @@
+:root {
+  --rp-steps-gap: 24px;
+  --rp-steps-size: 30px;
+}
+
 .rp-steps {
-  margin-left: 16px;
-  margin-bottom: 44px;
+  margin-left: 1rem;
+  margin-bottom: 2.5rem;
   border-left: 1px solid var(--rp-c-text-3);
-  padding-left: 24px;
+  padding-left: var(--rp-steps-gap);
   counter-reset: step;
 
   h3 {
     counter-increment: step;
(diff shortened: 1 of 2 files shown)


SoonIter:
thanks for your contribution ❤️️

I've gone through this PR and get what you're after. Right now, the look of the Steps component is kind of hideous and definitely calls for some optimization.  https://github.com/web-infra-dev/rspress/issues/1527


image


In Nextra's Steps component, all the anchor points are on the right, whereas in Rspress, Vitepress and such, they're on the left. It'd be best if we put the anchors on the left, just like for h1 - h4 headings

SoonIter (approved):
After my change

image

packages/core/src/theme/components/Steps/index.scss
```diff
@@ -1,51 +1,54 @@
...
-  padding-left: 24px;
+  padding-left: var(--rp-steps-gap);
   counter-reset: step;
 
   h3 {
     counter-increment: step;
-    margin-top: 3px;
+    margin-top: 0;
     position: relative;
+    display: flex;
```
JounQin, line 17: @SoonIter Is this required?

Without `display: flex`:

Image

With `display: flex`:

Image

Issue #3277, opened 2026-04-05
[Bug]: Page data is loaded async


elliotcourant:
### Version

```sh
System:
    OS: Linux 6.12 Debian GNU/Linux 13 (trixie) 13 (trixie)
    CPU: (32) x64 AMD Ryzen 9 7950X 16-Core Processor
    Memory: 48.10 GB / 61.95 GB
    Container: Yes
    Shell: 5.9 - /usr/bin/zsh
  Browsers:
    Chrome: 146.0.7680.177
    Firefox: 140.9.0esr
    Firefox Developer Edition: 140.9.0esr
  npmPackages:
    @rspress/config: workspace:* => 1.0.0
```

### Details

https://github.com/elliotcourant/rspress/blob/70ac277d23fe4a73f63c1945ca7232b7aab2359f/packages/core/src/runtime/App.tsx#L13-L30

When a page is initially loaded its already rendered by SSG, but if you then navigate to a page that relies on a hook of any form from the initial page you will get a very brief flash of the page being rendered incorrectly.

This is because anything using the page data won't have the most up to date data but the render is not blocked because the data is not undefined (since some data has already been loaded somewhere).

### Reproduce link

https://rspress.rs/api/

### Reproduce Steps

1. Navigate to `https://rspress.rs/api/`
2. Then click on the rspress logo in the top left hand corner to go to the home page

There will be an extremely brief flash of content that isn't the intended way the home page should be rendered.

You can make this more evident by setting the performance in chrome to be much slower than your computer as I've done in the example videos I've attached

---

### Nextra (baseline)

I'm using nextra as a baseline has as that is what I'm migrating from and how I noticed this bug:

https://github.com/user-attachments/assets/e97b0881-6dba-4c4b-9ae3-0ba19dfbed51

When nextra navigates between pages there is now flash of bad content

### Rspress (personal)

This is the rspress version of the exact same site

https://github.com/user-attachments/assets/99472cbc-5970-40b0-868a-94e54d2ee423

^ In this one you can see it render the new sidebar but the old table of contents remains for a moment. As well as rendering my custom inner component. Then once the frontmatter data is available the page resets back to the intended layout.

### Rspress (official)

This is the same behavior being shown with the current rspress site https://rspress.rs/api/

https://github.com/user-attachments/assets/958113bf-239d-46f0-9251-8b9adbe69095

[...]

PR #3278, opened 2026-04-05
fix(core): Fixed async page data loading

elliotcourant:
## Summary

Page data is now cached immediately after being lazy loaded, this way upon navigation we can try to set page data on the context using the information from the cache itself. This should remove a flash of content being odd because of stale pagedata caching.

This solution is a huge hack though, and I think I'd prefer to find a way to simply make this data available synchronously somehow. Or somehow tie it to the navigation in such a way that it is swapped in and out at the same time as the actual page content is.

### Before

https://github.com/user-attachments/assets/cea668e6-4fba-44f0-9dc6-defe81030081

### After

https://github.com/user-attachments/assets/ef90b20a-8f5c-4395-b453-1f5f13d4f4c3

## Related Issue

https://github.com/web-infra-dev/rspress/issues/3277

## Checklist

I haven't updated tests because I genuinely have no idea how I'd even test this. Happy to take suggestions

### DIFF
--- a/packages/core/src/runtime/index.ts
+++ b/packages/core/src/runtime/index.ts
@@ -26,7 +26,7 @@ export {
 export { useSite } from './hooks/useSite';
 export { useVersion } from './hooks/useVersion';
 export { useWindowSize } from './hooks/useWindowSize';
-export { initPageData } from './initPageData';
+export { initPageData, warmPageData } from './initPageData';
 export { NoSSR } from './NoSSR';
 export { isActive, pathnameToRouteService, preloadLink } from './route';
 export {
--- a/packages/core/src/runtime/index.ts
+++ b/packages/core/src/runtime/index.ts
@@ -26,6 +26,7 @@ export {
 export { useSite } from './hooks/useSite';
 export { useVersion } from './hooks/useVersion';
 export { useWindowSize } from './hooks/useWindowSize';
+export { initPageData } from './initPageData';
 export { NoSSR } from './NoSSR';
 export { isActive, pathnameToRouteService, preloadLink } from './route';
 export {
--- a/packages/core/src/runtime/initPageData.ts
+++ b/packages/core/src/runtime/initPageData.ts
@@ -121,7 +121,7 @@ export async function initPageData(routePath: string): Promise<Page> {
     } = MDX_OR_MD_REGEXP.test(matchedRoute.filePath)
       ? meta
       : (mod as unknown as PageMeta);
-    const result: Page = {
+    return {
       ...rest,
       pagePath,
       ...extractPageInfo,
@@ -130,7 +130,6 @@ export async function initPageData(routePath: string): Promise<Page> {
       frontmatter,
       toc,
     };
-    return result;
   }
 
   let lang = siteData.lang || '';
@@ -162,7 +161,7 @@ export async function initPageData(routePath: string): Promise<Page> {
   }
 
   // 404 Page
-  const result: Page = {
+  return {
     pagePath: '',
     pageType: '404',
     routePath: '/404',
@@ -174,5 +173,4 @@ export async function initPageData(routePath: string): Promise<Page> {
     _filepath: '',
     _relativePath: '',
   };
-  return result;
 }
--- a/packages/core/src/runtime/initPageData.ts
+++ b/packages/core/src/runtime/initPageData.ts
@@ -19,6 +19,12 @@ type PageMeta = {
 
 export type Page = PageDataLegacy['page'];
 
+const pageDataCache = new Map<string, Page>();
+
+export function getCachedPageData(routePath: string): Page | undefined {
+  return pageDataCache.get(routePath);
+}
+
 export async function initPageData(routePath: string): Promise<Page> {
   const matchedRoute = pathnameToRouteService(routePath);
   if (matchedRoute) {
@@ -51,7 +57,7 @@ export async function initPageData(routePath: string): Promise<Page> {
     } = MDX_OR_MD_REGEXP.test(matchedRoute.filePath)
       ? meta
       : (mod as unknown as PageMeta);
-    return {
+    const result: Page = {
       ...rest,
       pagePath,
       ...extractPageInfo,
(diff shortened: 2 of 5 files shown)


elliotcourant:
This fixes it for me locally, but this feels like a hack. I'm tinkering with a different way to do this which would make this not async at all. At least not anymore async than the lazy loading of page modules already is.

packages/core/src/runtime/initPageData.ts
```diff
@@ -19,6 +19,12 @@ type PageMeta = {
 
 export type Page = PageDataLegacy['page'];
 
+const pageDataCache = new Map<string, Page>();
```
JounQin: The cache is never cleared which could result high memory usage?

elliotcourant: Thats correct, for a small site this is probably fine-ish? but for a big site (say sentry's doc site as an example) that could result in a huge cache if the user is spending a substantial amount of time browsing it.

I could try and spend some more time on that and implement something where it keeps the last `n` pages worth of data, and then drops pages beyond that. However I believe the browser is going to keep the lazy-loaded modules of those pages in its own memory anyway since they only get evicted by true navigation. So I'm not sure how much of a difference that would make unless there is a mechanism for throwing out modules that arent being used already?

packages/core/src/runtime/initPageData.ts
```diff
@@ -103,4 +111,6 @@ export async function initPageData(routePath: string): Promise<Page> {
     _filepath: '',
     _relativePath: '',
   };
+  pageDataCache.set(routePath, result);
```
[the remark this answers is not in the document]
elliotcourant: Removed this!

packages/core/src/runtime/index.ts
```diff
@@ -26,6 +26,7 @@ export {
 export { useSite } from './hooks/useSite';
 export { useVersion } from './hooks/useVersion';
 export { useWindowSize } from './hooks/useWindowSize';
+export { initPageData } from './initPageData';
```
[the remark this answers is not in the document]
elliotcourant: I changed this slightly so that init page data doesnt import from runtime anymore. It just imports from the files directly. This way its no longer circular.

packages/core/src/runtime/initPageData.ts
elliotcourant: Sorry if that sounded like I'm rambling.

Let me try to noodle on some other ways to do this, maybe I only need to cache 3 items at most for example. The first item being the potential next page (whichever one the user hovers, swap it when the module is loaded), the second being the current page, and the last being the previous page.

Forward navigation is just validating that the first item in the array is defined and is for the same route, then shifting the array over one and throwing away the last most item.

Backward navigation is doing the same thing in reverse, but that way the previous page is always in the cache.

This way there are at most 3 pages worth of data in the cache and it should be sufficient for most navigation. Faster navigation will be limited by the lazy module loading anyway so this will be more bounded.

elliotcourant: Okay I re-implemented this as an array of 3 items. This way the cache is completely bounded, and it moves items into the cache based on navigation forwards or backwards. I've commented the code heavily to try to make it clear.

packages/core/src/theme/components/Link/useLinkNavigate.ts
```diff
@@ -89,6 +90,7 @@ export function useLinkNavigate(
               nprogress.start();
             }, 200);
             await matchedRoute.preload();
```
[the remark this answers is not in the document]
elliotcourant: Updated!

packages/core/src/runtime/initPageData.ts
```diff
@@ -91,7 +162,7 @@ export async function initPageData(routePath: string): Promise<Page> {
   }
 
   // 404 Page
-  return {
+  const result: Page = {
```
elliotcourant: These changes aren't needed anymore, I'll clean them up tomorrow

packages/core/src/runtime/initPageData.ts
```diff
@@ -51,7 +121,7 @@ export async function initPageData(routePath: string): Promise<Page> {
     } = MDX_OR_MD_REGEXP.test(matchedRoute.filePath)
       ? meta
       : (mod as unknown as PageMeta);
-    return {
+    const result: Page = {
```
elliotcourant: Ditto this change, not needed now

SoonIter:
Thanks for raising this ❤️ You've identified a real pain point here — exactly the kind of thing experienced front-end devs notice :D

While this fix works, I'm wondering if we should address it at the root. I'm exploring a refactor to React Router's data mode, which would handle data fetching more cleanly and likely resolve this class of issues.

Would love your thoughts on that approach

SoonIter (approved):
merged and will be released in 2.0.9

elliotcourant:
> Thanks for raising this ❤️ You've identified a real pain point here — exactly the kind of thing experienced front-end devs notice :D
> 
> While this fix works, I'm wondering if we should address it at the root. I'm exploring a refactor to React Router's data mode, which would handle data fetching more cleanly and likely resolve this class of issues.
> 
> Would love your thoughts on that approach

I'm very unfamiliar with the magic that can be done with react-router, however from the exploring I did trying to find other ways of making this work its very tricky. I'd be curious to see what you had in mind with the data mode and how that works?

Ideally react-router _should_ handle all of this and do it via a transition such that all of the promises for the new page being loaded get resolved before the swap of the DOM takes place (I thought suspense + some react-router knobs and dials allowed for this behavior?). But getting it to handle that all in one go isn't straightforward to my knowledge?

I had explored trying to embed the data that would be loaded in the lazy module itself as well and instead of using context to pass the data around instead it would always be self-referencing with whichever module was currently loaded and being displayed. But this broke so many other things with how I wanted to make it work that it didn't seem worth it.

I also have no idea how nextra + nextjs handles this

elliotcourant:
Also I just realized I haven't been using changeset to add release notes for these PRs, I will do that in the future sorry!

SoonIter:
> Also I just realized I haven't been using changeset to add release notes for these PRs, I will do that in the future sorry!

Don't worry, this repo's changeset is useless. I'll remove the entire changeset tool from the Rspress repository, and this isn't just done for you. :D

PR #3297, opened 2026-04-08
perf(deps): bundle `gray-matter` to reduce install size and deps complexity

SoonIter:
This PR reduces published install footprint by bundling `gray-matter` through `@rspress/shared` instead of shipping it as a runtime dependency. It also updates the branch onto the latest `origin/main` so the change composes with the newer `@rspress/shared` bundled exports.

- **Packaging**
  - moves `gray-matter` from `dependencies` to `devDependencies` in `packages/shared`
  - adds a dedicated `@rspress/shared/gray-matter` subpath export
  - keeps the package publish surface aligned with existing bundled helper exports

- **Shared wrapper**
  - adds `src/grayMatter.ts` as the bundling entry
  - preserves the upstream call shape needed by internal callers, including generics and optional parser options
  - avoids the DTS extraction issue that comes from exporting `gray-matter` directly

- **Internal usage**
  - switches `loadFrontMatter` to consume the shared wrapper instead of the external package directly
  - lets the frontmatter generic flow through the parser call without extra casting at the call site

- **Mainline sync**
  - merges the latest `origin/main`
  - resolves overlap with the newer bundled `github-slugger` export so both bundled entries coexist in `@rspress/shared`

```ts
import { grayMatter } from '@rspress/shared/gray-matter';

const { content, data } = grayMatter<{ title: string }>(source, options);
```


SoonIter:
@copilot+gpt-5.4 rebase to latest origin/main

Timeless0911 (approved):

PR #3306, opened 2026-04-13
feat(core/router): migrate to react-router data mode and support react-router-dom@6

SoonIter:
## Summary

This PR migrates the Rspress core runtime to React Router data mode and adds compatibility for "react-router-dom@6". It also introduces support for versioned and localized 404 routes.

## Related Issue

None

## Checklist


SoonIter:
Superseded by #3685, which migrates navigation and HTML/Markdown static rendering to React Router Data mode on current main. Page loading now runs in a router loader while retaining Rspress' existing URL matching rules.

Closing this older migration approach. This is not a claim that every change here was ported: React Router 6 server-import compatibility remains a separate follow-up in #3125, and the generated localized/versioned fallback 404 routes plus associated plugin changes are not included in #3685.

PR #3357, opened 2026-05-07
feat(mdx/FileCodeblock): support file codeblock starts with / prefix

### LINKED ISSUE #3336: [Feature]: use `/` prefix to point to `config.root` instead of `<root>/` for `cwd` for code block filepath

It's weird that `<root>/` points to `cwd` instead of `config.root` because there is often `docs` directory for `config.root`, if we have shared files at `docs/assets/xyz.sh`, and the absolute path must be written as `<root>/docs/assets/xyz.sh`, ideally it could be `/assets/xyz.sh`.

Only `./`, `../` and `<root>/` prefixes are allowed currently, `/` prefix support could be added as `config.root` alias.

JounQin:
## Summary

support file codeblock starts with `/` prefix

## Related Issue

close #3336

## Checklist

### DIFF
--- a/packages/core/src/node/mdx/options.ts
+++ b/packages/core/src/node/mdx/options.ts
@@ -75,7 +75,7 @@ export async function createMDXOptions(options: {
       cjkFriendlyEmphasis && remarkCjkFriendlyGfmStrikethrough,
       remarkToc,
       !isSsgMd && remarkContainerSyntax,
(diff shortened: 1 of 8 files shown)


SoonIter (approved):
Thanks ❤️️~

The `root` configuration name is confusing. I checked Docusaurus, but I couldn't find a similar configuration to achieve this.

PR #3391, opened 2026-05-15
feat(theme/lastUpdated): support displaying `lastUpdated.author`

### LINKED ISSUE #725: [Feature]: Support `lastUpdatedAuthor`

It is nice that `rspress` supports [`lastUpdated`](https://rspress.dev/api/config/config-theme.html#lastupdated)! But popular tools like docusaurus supports things like: [`showLastUpdateAuthor`](https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs#showLastUpdateAuthor) which shows not only the updated time but also the author.

E.g.: 
image

It would be nice if `rspress` could support this. It would be even better if the displayed content could be customized(we use `pnpm patch` tricks in docusaurus to display the `user.email` of the author instead of the default `user.name`).

- Enabled by `lastUpdated: true`.
- New config: `lastUpdatedAuthor`:
  - Set to `false` to disable displaying author.
  - Set to `(path: string) => string` to modify the display text.

natefzz:
## Summary

Adds a new `lastUpdatedAuthor` theme config to display the last commit author alongside the last updated time on doc pages.

When `lastUpdated: true` is enabled:

- **default**: also display the commit author's name (`Last Updated: ... by Alice`).
- `lastUpdatedAuthor: false`: hide the author, keep the timestamp as before.
- `lastUpdatedAuthor: ({ name, email, filePath }) => string`: build-time callback to customize the displayed text (e.g. show the email instead of the name). `filePath` is the source file of the doc, useful for path-based formatting.

The author info comes from the same `git log` call already used for the timestamp (extended from `--format=%at` to `--format=%at%n%an%n%ae`), so there is no extra subprocess per page.

### Notes for reviewers

- Behavior change: existing users with `lastUpdated: true` will now see the commit author appear after the timestamp. The author can be hidden with `lastUpdatedAuthor: false` if undesired.
- i18n: added `lastUpdatedAuthorText` to the five default locales (en/zh/ja/ko/ru).

## Related Issue

Closes #725

## Checklist

- [x] Tests updated (or not required). *no existing tests for the `lastUpdated` plugin; manually verified in `website` dev server*
- [x] Documentation updated (or not required). *updated `website/docs/{en,zh}/ui/layout-components/last-updated.mdx`*

### DIFF
--- a/packages/core/src/node/PluginDriver.ts
+++ b/packages/core/src/node/PluginDriver.ts
@@ -68,7 +68,7 @@ export class PluginDriver {
     const mediumZoomConfig = config?.mediumZoom ?? true;
     if (enableLastUpdated) {
       const { pluginLastUpdated } = await import('./last-updated/index');
-      this.addPlugin(pluginLastUpdated());
+      this.addPlugin(pluginLastUpdated(themeConfig?.lastUpdatedAuthor));
     }
     if (mediumZoomConfig) {
       const { pluginMediumZoom } = await import('./medium-zoom/index');
--- a/packages/core/src/node/PluginDriver.ts
+++ b/packages/core/src/node/PluginDriver.ts
@@ -64,11 +64,17 @@ export class PluginDriver {
     this.clearPlugins();
     const config = this.#config;
     const themeConfig = config?.themeConfig || {};
-    const enableLastUpdated = themeConfig?.lastUpdated;
+    const lastUpdatedConfig = themeConfig?.lastUpdated;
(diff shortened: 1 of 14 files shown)


SoonIter:
Thanks for your contribution! This is a useful improvement and helps move #725 forward. 👍🏻

I made a few adjustments to keep `lastUpdated: true` backward-compatible and expose the new behavior through `lastUpdated.author`. I also updated the related tests, types, i18n text, and docs.

Thanks again!

SoonIter (approved):

PR #3405, opened 2026-05-21
fix(ssg): remove unused boolean define for import.meta.env.SSR after Rsbuild 2.0.7

elecmonkey:
## Summary

Since Rsbuild 2.0.7, `import.meta.env.SSR` is expected to use boolean semantics by https://github.com/web-infra-dev/rsbuild/pull/7700.

Rspress still declares `@rsbuild/core` as `^2.0.6` now, but other repo's dependency resolution may install 2.0.7+ when other packages need a newer compatible version, cause a DefinePlugin warning like:

```
Conflicting values for 'import.meta.env.SSR' (false !== "false")
```


packages/core/src/node/initRsbuild.ts
```diff
@@ -425,7 +425,7 @@ async function createInternalBuildConfig(
           define: {
             'process.env.__SSR__': JSON.stringify(false),
             'process.env.__SSR_MD__': JSON.stringify(false),
-            'import.meta.env.SSR': JSON.stringify(false),
```
SoonIter, line 428: If we upgrade to Rsbuild 2.0.7, we could delete this line directly, right?

So upgrade to 2.0.7 is a better choice

SoonIter: ^2.0.6 -> ^2.0.7

SoonIter:
@copilot+gpt-5.4
rebase to the latest origin/main and adopt the review suggestions

SoonIter:
@copilot+gpt-5.4 

Fix `Conflicting values for 'import.meta.env.SSR' (false !== "false")` issue

SoonIter (approved):

PR #3413, opened 2026-05-23
fix(plugin-algolia): Add `react-dom` `react` peerDependencies and optimize Search component preconnect ⚡

sanjaiyan-dev:
## Summary
This PR replaces the manual DOM insertion of the Algolia preconnect `<link>` tag inside `useEffect` with React's native [`preconnect` API from `react-dom`](https://react.dev/reference/react-dom/preconnect).

By adopting the native `react-dom` `preconnect` API, we achieve several benefits:
* **Earlier Initialization:** The resource hint is registered during the render phase, allowing the browser to begin resolving DNS and establishing the connection sooner.
* **SSG Support:** If the page is server-rendered, the `<link rel="preconnect">` tag is injected directly into the initial HTML stream, making it discoverable prior to client-side JS execution.
* **Automatic Deduplication:** React handles duplicate calls to the same origin, avoiding redundant link injections if the component mounts multiple times or re-renders.
* **Code Simplification:** Removes manual imperative DOM creation, cleanup logic, and effect dependencies.

#### Before (useEffect Approach)

```
Server HTML (No Tag) ──> Downloads JS Bundle ──> React Boots Up ──> Component Paints ──> useEffect Fires ──> <link> Appended ──> Preconnect Starts (LATE)
``` 

#### After With Static-Site Generation (SSG):

```
Build Time (Static HTML built WITH tag) ──> Browser Loads HTML ──> Preconnect Starts Immediately (IMMEDIATE)
                                                      └──> Downloads JS Bundle / Hydrates (Parallel)
``` 

## Checklist


SoonIter:
Thanks for the PR!

The optimization makes sense, but Rspress currently supports both React 18 and React 19. Since the `preconnect` API from `react-dom` is only available in newer React versions, we need to handle the compatibility with React 18 before merging this change.

I’ll make some adjustments when I have time. Thanks again for looking into this!

sanjaiyan-dev:
> Thanks for the PR!
> 
> The optimization makes sense, but Rspress currently supports both React 18 and React 19. Since the `preconnect` API from `react-dom` is only available in newer React versions, we need to handle the compatibility with React 18 before merging this change.
> 
> I’ll make some adjustments when I have time. Thanks again for looking into this!

Hi @SoonIter,

Thanks for the review! That makes complete sense. To ensure compatibility across both React 18 and React 19, we can use runtime feature detection on the `react-dom` module. 

If the environment is running React 19+, we can dynamically resolve and call `preconnect` during the render phase. If it is running React 18, we can safely fall back to the original `useEffect` implementation. 

By using a namespace import (`import * as ReactDOM`) and checking the existence of the method, we also prevent bundler errors on older versions that do not export the named `preconnect` binding.

Here is the proposed implementation:

```tsx
import React, { useEffect } from 'react';
import * as ReactDOM from 'react-dom';

// ... inside the Search component ...

const safePreconnect = (ReactDOM as any).preconnect;

if (typeof safePreconnect === 'function') {
  // React 19 path (runs during render / SSG)
  safePreconnect(algoliaUrl, { crossOrigin: 'anonymous' });
}

useEffect(() => {
  // React 18 fallback (runs client-side only)
  if (typeof safePreconnect === 'function') {
    return;
  }

  const link = document.createElement('link');
  link.rel = 'preconnect';
  link.href = algoliaUrl;
  link.crossOrigin = '';
  document.head.appendChild(link);

  return () => {
    document.head.removeChild(link);
  };
}, [algoliaUrl]);

SoonIter (approved):
LGTM ❤️️

PR #3425, opened 2026-05-28
feat(theme): Add React DOM-based preloading to Image component 🖼️

sanjaiyan-dev:
#### Overview
This PR introduces support for `ReactDOM.preload` in our custom `Image` component to optimize image loading. By utilizing the React 19 resource loading APIs, this component can trigger preloads that are automatically deduplicated and injected into the document head (including during Static-Site Generation).

#### Key Changes
- **New `Image` Component**: Created a wrapper around the native HTML `` tag that accepts a `preload` boolean prop.
- **`ReactDOM.preload` Integration**: Safely checks for and invokes `ReactDOM.preload` (referencing [React DOM preload documentation](https://react.dev/reference/react-dom/preload)) to preload images. It handles configuration parameters like `imageSrcSet`, `imageSizes`, `fetchPriority`, `crossOrigin`, and `referrerPolicy`.
- **MDX Component Integration**: (Optional / context-specific) Set up MDX-based image tags to trigger a preload automatically if the image is configured with `loading="eager"` and `fetchPriority` is _not_ set to `'low'`.

#### Why this is helpful
Preloading critical assets early in the loading cycle helps improve Web Vital metrics, particularly Largest Contentful Paint (LCP). 

This is particularly useful for:
* Hero images (e.g., `HomeHero`)
* Navigation bar logos or primary icons
* Above-the-fold content in blog posts and documentation

## Usage Example

```tsx
import { Image } from '@rspress/core/theme';

// Manual preload
<Image src="/hero.png" preload />

```

```mdx
{/* Auto-preload via eager loading props in mdx */}
<img src="/blog-banner.png" loading="eager" />
```

> Please let me know if there are any errors or if the facts can be improved. Any feedback on the documentation or implementation details is highly appreciated.

## Checklist


sanjaiyan-dev:
Hi @SoonIter,

Apologies for the ping. I have updated the `Image` component to resolve `ReactDOM.preload` safely at runtime. This prevents the previous build and dev-server crashes in React 18 environments while still enabling the preloading optimization in React 19.

The CI/CD checks should now pass successfully. Please let me know if this is ready for your review and merge. 

Thank you!

SoonIter (approved):
Thanks for your contribution! ❤️️ This is a great improvement. 

I’ll merge it first, and after it lands on `main`, I’ll follow up with the documentation updates and any necessary adjustments.

SoonIter:
I think we may want to reconsider this change for now.

From what I observed, React 19 already generates resource hints for eligible SSR/SSG-rendered `` elements. In Rspress output, `<link rel="preload" as="image">` is already emitted even without explicitly calling `ReactDOM.preload`.

image


Because of that, manually calling `ReactDOM.preload` from an `Image` wrapper seems to overlap with React DOM's built-in behavior for the main initial page-load path. Since React already handles the preload injection and deduplication logic, we may not need to maintain a separate layer for this in Rspress.

sanjaiyan-dev:
> I think we may want to reconsider this change for now.
> 
> From what I observed, React 19 already generates resource hints for eligible SSR/SSG-rendered `` elements. In Rspress output, `<link rel="preload" as="image">` is already emitted even without explicitly calling `ReactDOM.preload`.
> 
> image
> Because of that, manually calling `ReactDOM.preload` from an `Image` wrapper seems to overlap with React DOM's built-in behavior for the main initial page-load path. Since React already handles the preload injection and deduplication logic, we may not need to maintain a separate layer for this in Rspress.



Hi @SoonIter ,

Thank you for the explanation! I wasn't fully aware of React 19's automatic preloading behavior, so this is very helpful to learn. 

As a reference, Next.js uses `ReactDOM.preload` in its [Image component](https://github.com/vercel/next.js/blob/0e6f52e99b5d161ebd02f46c4b5a8ed520ba0716/packages/next/src/client/image-component.tsx#L345) to give developers explicit control. A manual opt-in can help avoid the performance overhead of preloading too many assets, while also allowing precise preloading for responsive images using `imageSrcSet` and `imageSizes`.

I apologise for any inconvenience.

Issue #2325, opened 2025-06-26
[Refactor]: refactor `mediumZoom` to `<Img />` in `getCustomMDXComponent`


SoonIter:
### What problem does this feature solve?

a lot of bugs to solve in medium-zoom usage

### What does the proposed API look like?

1. remove `@rspress/plugin-medium-zoom`

2. should refactor it to `Img` in `getCustomMDXComponent`

Related reference info:

- https://github.com/web-infra-dev/rspress/blob/ad5fac97d4a63069dd68d50d0715ce26b28b8d93/packages/theme-default/src/layout/DocLayout/docComponents/img.tsx#L5

- Good design in fumadocs: https://github.com/fuma-nama/fumadocs/blob/5723bbe58ef805a5421a780abf235a10b251be2f/packages/ui/src/components/image-zoom.tsx#L6

Tsukistar:
excuse me, I have a question for the issue.
should all the `Img` zoomable after refactor, or the user also need to configure the `selector` in the `mediumZoom` in `rspress.config.ts`

SoonIter:
> excuse me, I have a question for the issue. should all the `Img` zoomable after refactor, or the user also need to configure the `selector` in the `mediumZoom` in `rspress.config.ts`

sorry for the late reply

it doesn't matter, it can be seen as a breaking change

Malusnow:
hi, @SoonIter ,I've already moved the Medium Zoom logic to the Img component. Next, I plan to completely remove the old global MediumZoom plugin chain, related configuration items, and documentation, so that all default MDX images should be scalable by default in the new Img component. I want to confirm if this direction is correct, or whether I should retain the optional scaling configuration for the img element.

PR #3437, opened 2026-06-03
refactor(theme-default): move medium zoom into Img component

Malusnow:
## Related Issue

#2325
## Checklist


SoonIter:
Sorry for taking so long to get back to you, and thank you for the contribution.

We have been considering related image features since before the v2 release. However, given the existing `medium-zoom` configuration, this change would be breaking. We do not plan to move forward with this PR for now; we would prefer to revisit it as part of a broader effort around image handling.

One major issue in Rspress today is image-related CLS: users still need to specify image height and width explicitly. This is something we would like to address together with the broader image work. For reference, Fumadocs has a relevant implementation here: [remark-image](https://www.fumadocs.dev/docs/headless/mdx/remark-image).

PR #3475, opened 2026-06-24
refactor(theme): use `useLayoutEffect` in useNavTransparent for performance🛠️

sanjaiyan-dev:
### Description
This change replaces `useEffect` with `useLayoutEffect` in the scroll-tracking class toggle for the navigation bar.

### Why is this change necessary?
Using `useEffect` causes the class manipulation on `document.body` to run asynchronously after the browser has already painted the frame. This results in a temporary "flash of unstyled content" (FOUC) or visual flicker, where the navigation bar is briefly drawn with the wrong background state immediately after crossing the scroll threshold. 

### How does this address the issue?
Replacing `useEffect` with `useLayoutEffect` ensures that the class toggle happens synchronously after the scroll state changes, but before the browser paints the updated layout. This binds the scroll threshold crossing and the style change to the same rendering frame, removing the visual delay.

### Key Changes
* Switched the `body.classList` updater logic from `useEffect` to `useLayoutEffect` in the navigation style hook.

When a component updates, the browser must process a visual pipeline:  
$$\text{DOM Changes} \longrightarrow \text{Recalculate Styles} \longrightarrow \text{Layout (Reflow)} \longrightarrow \text{Paint} \longrightarrow \text{Composite}$$

By intercepting the render pipeline before the browser draws pixels, useLayoutEffect _halves the rendering_ workload for that frame.


SoonIter (approved):
LGTM, thanks! ❤️️

PR #3476, opened 2026-06-24
refactor(router): `await` navigate and opt-in to router `useTransition` 🧭

sanjaiyan-dev:
## Summary

This PR addresses navigation transition behavior. By adding the missing `await` to the `navigate` call and enabling the `useTransition` flag, we ensure that navigations are fully and correctly coordinated with React's transition system.

### Key Changes
* **Awaited Navigation Promise:** Added `await` to `navigate(removeBaseHref, ...)` inside `preloadChunkThenNavigate`. Per the [React Router transitions documentation](https://reactrouter.com/explanation/react-transitions#opt-in-via-usetransitionstrue), the transition will end prematurely if the navigation promise is not returned or awaited.
   >[Important: You **must** always `return` or `await` the `navigate` promise inside `startTransition` so that the Transition encompasses the full duration of the navigation. If you forget to `return` or `await` the promise, the Transition will end prematurely and things won't work as expected.](https://reactrouter.com/explanation/react-transitions#opt-in-via-usetransitionstrue)
* **Enabled `useTransition={true}`:** Added this flag to the router configuration. This opts the application into React-Router's Transition-enabled navigation mode, helping future-proof the codebase and allowing link/form transitions to remain active for the full duration of navigation tasks. 

### References
* [React Router - Transitions Opt-In](https://reactrouter.com/explanation/react-transitions#opt-in-via-usetransitionstrue)
## Checklist


sanjaiyan-dev:
Hi @SoonIter 
I am very sorry to trouble you and take up your time.
Whenever you have a moment to spare, would it be possible to review this pull request?

packages/core/src/runtime/ClientApp.tsx
```diff
@@ -30,7 +30,10 @@ export function ClientApp({
       <PageContext.Provider
         value={useMemo(() => ({ data, setData }), [data, setData])}
       >
-        <BrowserRouter basename={removeTrailingSlash(withBase('/'))}>
+        <BrowserRouter
+          basename={removeTrailingSlash(withBase('/'))}
+          useTransitions={true}
```
SoonIter, line 35: Image

<del>We seem to be using react-router v7, and this option is already enabled by default.</del>

Sorry, I misunderstood

Can I get the before-and-after differences to test this behavior?

sanjaiyan-dev: > Image
> ~We seem to be using react-router v7, and this option is already enabled by default.~
> 
> Sorry, I misunderstood
> 
> Can I get the before-and-after differences to test this behavior?


Hi @SoonIter,

Actually, there won't be any immediate visual behavior changes from adding `useTransitions={true}` right now, as we don't currently use React Router's native `Link` component in our codebase. I included it mainly as a way to future-proof the routing configuration for the codebase.

The primary fix in this PR is adding the missing `await` to the `navigate` calls. Without it, the transition ends prematurely, which is a known issue highlighted in the official React Router documentation (which explicitly states that the navigation promise must be returned or awaited inside `startTransition` to work correctly). 

Sorry for any confusion, and let me know if this explanation makes sense or if you have any other questions!

sanjaiyan-dev: Screenshot_2026-07-01-21-27-59-77_40deb401b9ffe8e1df2f1cc5ba480b12

SoonIter (approved):
LGTM ❤️️~

thanks for your explanation

sanjaiyan-dev:
> LGTM ❤️️~
> 
> thanks for your explanation

Hi @SoonIter,
Thanks again for reviewing and merging this PR!
I apologize for bringing up another request so soon after you just merged this one. I have submitted a follow-up PR to address Markdown and MDX links here:  #3479.

PR #3479, opened 2026-06-25
feat(theme/Link): Add `route.useTransitions` to opt-out startTransition navigation

sanjaiyan-dev:
## Summary

This pull request introduces a new configuration option, `concurrentRoute`, to enable `startTransition`-based routing for links parsed from markdown and MDX files. 

### Why this approach?
Authors writing content in `.md` or `.mdx` files use standard markdown link syntax (e.g., `[Text](/path)`). Because they cannot directly attach React props (such as `startTransition`) to these elements, enabling concurrent routing on a per-link basis is not feasible for content creators. 

To solve this, we introduce a global configuration option. By reading this configuration at runtime, the internal link component (`A`) can conditionally apply `startTransition` during navigation.

## Checklist


SoonIter:
I pushed two follow-up updates with some code changes: the concurrent routing switch is now handled in the shared Link navigation path, and `route.concurrentRoute` is defaulted through site data, together with the related docs/type updates.

Before we decide how much opt-out surface we should keep here, have we already seen any concrete user case where the transition behavior / `useTransition` must be disabled? I want to confirm whether there is a real user opt-out requirement, or if the default behavior is sufficient for now.

SoonIter (approved):

PR #3501, opened 2026-07-07
fix(theme/codeblock): avoid list markers when copying code blocks

### LINKED ISSUE #3500: [Bug]: Code block copy includes parent list marker when code block is nested in list item
### Version

```sh
System:
    OS: macOS 15.7.4
    CPU: (16) arm64 Apple M4 Max
    Memory: 5.63 GB / 64.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Firefox: 152.0.4
    Safari: 18.6
  npmPackages:
    @rspress/core: ^2.0.6 => 2.0.14
    @rspress/plugin-client-redirects: ^2.0.14 => 2.0.14
    @rspress/plugin-sitemap: ^2.0.14 => 2.0.14
```

### Details

Copying a codeblock nested in a markdown bullet prepends the markdown list marker to the copied code e.g.:
- Nested example 
    ```
       echo foo
    ```
is copied as "- echo foo"

This appears to be a regression possible introduced here: https://github.com/web-infra-dev/rspress/pull/3333
But definitely introduced since the @rspress/core/dist/theme/logic/getCopyableText.js file was added @SoonIter 

I was able to reproduce in a sandbox, see below.

### Reproduce link

https://codesandbox.io/p/devbox/inspiring-saha-sqsypj?workspaceId=ws_UxU85dszBQxpjPWC7Aq6cx

### Reproduce Steps

- Run sandbox and open index page
- copy code blocks
- observe md list marker being prepended to copied code

maikthomas:
## Summary

This PR fixes code-copy behavior for code blocks nested in list items.

Previously, copying a code block inside an ordered or unordered list could prepend list markers (for example, `1. ` or `- `) to the copied text. This happened because markdown list formatting was applied during text extraction, even for code content.

This change updates the copy text extraction logic to skip markdown list marker rendering when the target is inside `pre` or `code`, so copied output is the exact code users expect.

Also included:
- Added unit tests for `getCopyableText` to cover:
  - code blocks inside ordered lists (no numeric prefix),
  - code blocks inside unordered lists (no dash prefix),
  - preservation of list markers for non-code list content.

## Related Issue

fixes https://github.com/web-infra-dev/rspress/issues/3500

## Checklist

### DIFF
--- a/packages/core/src/theme/logic/getCopyableText.test.ts
+++ b/packages/core/src/theme/logic/getCopyableText.test.ts
@@ -1,77 +1,259 @@
+import { afterEach, beforeEach, describe, expect, it } from '@rstest/core';
+
 import { getCopyableText } from './getCopyableText';
-import { describe, it, expect, beforeEach } from 'vitest';
+
+const ELEMENT_NODE = 1;
+const TEXT_NODE = 3;
+const SHOW_TEXT = 4;
+
+type FakeNode = FakeElement | FakeTextNode;
+
+interface FakeTextNode {
+  nodeType: number;
+  nodeName: '#text';
+  nodeValue: string;
+  parentElement: FakeElement;
+  previousSibling: FakeNode | null;
+}
+
+interface FakeElement {
+  nodeType: number;
+  nodeName: string;
+  tagName: string;
+  ownerDocument: FakeDocument;
+  parentElement: FakeElement | null;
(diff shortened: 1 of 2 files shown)


SoonIter (approved):

SoonIter:
LGTM, thanks for the fix!

PR #3503, opened 2026-07-07
fix(theme/Link): `startTransition` in async behaviour 🔧 

sanjaiyan-dev:
## Summary

I wanted to share some context on why I wrapped the `navigate` call in an additional `startTransition` block. 

According to the official React documentation on [Troubleshooting: React doesn't treat my state update after await as a Transition](https://react.dev/reference/react/useTransition#react-doesnt-treat-my-state-update-after-await-as-a-transition):

> *"When you use `await` inside a `startTransition` function, the state updates that happen after the `await` are not marked as Transitions. You must wrap state updates after each `await` in a `startTransition` call."*

Essentially, when an asynchronous operation is awaited, the subsequent code loses the active transition context. To ensure our `navigate` state update remains non-blocking, we need to wrap it like this:

```javascript
startTransition(async () => {
  await preloadChunkThenNavigate();
  
  // Wrapped again to preserve the transition context after the await
  startTransition(() => {
    return navigate('/new-path');
  });
});
```

<details><summary>In <code>useLinkNavigate.ts</code>, <code>preloadChunkThenNavigate</code> is initially wrapped in a transition. However, on line 98, an asynchronous chunk preloading operation occurs:</summary>
<p>

https://github.com/web-infra-dev/rspress/blob/45b87c9eb595436d4d49974085fffedceb1df03b/packages/core/src/theme/components/Link/useLinkNavigate.ts#L98

</p>
</details> 

Please let me know if you have any thoughts on this approach or if you would like me to adjust anything! 🤝

## Checklist


SoonIter (approved):
thanks~

PR #3508, opened 2026-07-09
feat(plugin-algolia): Integrate `startTransition` for navigation and refactor `preconnect` logic 🔍 

sanjaiyan-dev:
## Summary

This change introduces support for React concurrent navigation within the Algolia DocSearch component. #3479 

* **What was changed**: The navigation callback in the `<DocSearch>` navigator now conditionally wraps `navigate(itemUrl)` in a React `startTransition` block.
* **Why**: When the `useTransitions` flag is enabled, wrapping navigation in `startTransition` allows React to handle the page transition concurrently. This helps prevent the UI from blocking or freezing during heavy page loads after selecting a search result.

## Checklist


packages/plugin-algolia/src/runtime/Search.tsx
```diff
@@ -1,10 +1,15 @@
 import type { DocSearchProps } from '@docsearch/react';
 import { DocSearch } from '@docsearch/react';
-import { removeBase, useLang, useNavigate } from '@rspress/core/runtime';
+import {
+  removeBase,
+  useLang,
+  useNavigate,
+  useSite,
+} from '@rspress/core/runtime';
 import { Link } from '@rspress/core/theme';
```
SoonIter: I reckon it'd be a better call to use `useLinkNavigate` right here. useLinkNavigate has already wrapped up the logic regarding startTransition.

sanjaiyan-dev: > I reckon it'd be a better call to use `useLinkNavigate` right here. useLinkNavigate has already wrapped up the logic regarding startTransition.

Makes sense, I'll update it to use `useLinkNavigate`. Thanks!

sanjaiyan-dev: Hi @SoonIter,
I hope this is now correct 🤝

SoonIter (approved):

SoonIter:
thanks ❤️️~

PR #3518, opened 2026-07-14
perf(runtime): preload current route chunks with link tag

SoonIter:
## Summary

This PR adds the following tag to each generated page so its current async route chunk starts downloading immediately:

```html
<link rel="preload" href="/static/js/async/route-<hash>.<contenthash>.js" as="script">
```

This removes the extra network waterfall where the route code was only requested after the client runtime started hydration. The preload URL respects `base`, CDN asset prefixes, and `assetPrefix: 'auto'`.

## Related Issue

N/A

## Checklist


SoonIter:
300ms perf

## Before


image

## After

image

Timeless0911 (approved):

PR #3569, opened 2026-07-31
test(e2e): migrate to @rstest/playwright

9aoy:
## Summary

This PR migrates Rspress end-to-end tests from `@playwright/test` to `@rstest/playwright` to consolidate the test stack on Rstest while preserving the existing Chrome-based CI behavior.

It adds shared Rstest Playwright configuration, externalizes `@rspress/core` to preserve workspace runtime resolution, adapts the affected fixtures to Rstest lifecycle and assertion APIs, and uploads failure traces from macOS and Windows CI.

## Checklist


e2e/fixtures/basic/index.test.ts
```diff
@@ -115,11 +119,11 @@ test.describe('themeConfig.darkMode', async () => {
       waitUntil: 'networkidle',
     });
 
-    await expect.poll(() => getIsDark(page)).toBe(true);
+    await expect.poll(() => getIsDark(page), { timeout: 5000 }).toBe(true);
```
SoonIter: could you explain why we need to set 5000ms timeout?

e2e/fixtures/hmr/index.test.ts
```diff
@@ -32,6 +32,17 @@ test.describe('hmr test', async () => {
 
   const getRestartCount = () =>
     devOutput.match(/restarting server as .* changed/g)?.length ?? 0;
+  const getBuildCount = () => devOutput.match(/built in/g)?.length ?? 0;
```
SoonIter, line 36: could you explain what's the difference from before?

e2e/fixtures/banner/index.test.ts
```diff
@@ -1,4 +1,4 @@
-import { expect, test } from '@playwright/test';
+import { expect, test } from '@e2e/test';
```
SoonIter, line 1:
```suggestion
import { expect, test } from '@rstest/core';
```
Is it possible for us to import directly from @rstest/core?

SoonIter: or @rstest/playwright

e2e/fixtures/basic/index.test.ts
9aoy: Playwright's default `expect` timeout is 5 seconds, while Rstest's `expect.poll` defaults to 1 second. Setting it explicitly preserves the previous behavior and avoids making this asynchronous UI update more flaky after the migration.

e2e/fixtures/hmr/index.test.ts
9aoy: Previously the test continued as soon as the restart log appeared, although the rebuild could still be running. Consecutive file changes could therefore overlap or be coalesced. The new helper waits for both the restart event and a subsequent completed build, and the `finally` block makes retries deterministic by restoring the fixture state.

e2e/fixtures/banner/index.test.ts
9aoy: `@rstest/core` does not provide the `page` fixture. Importing directly from `@rstest/playwright` would work, but it would bypass our shared fixture that configures the CI Chrome channel, viewport, and `on-first-retry` trace behavior. `@e2e/test` keeps those options consistent across all browser tests.

SoonIter (approved):

SoonIter:
LGTM ❤️️

PR #3586, opened 2026-08-05
perf(core): enable React Optimizer for runtime and theme

sanjaiyan-dev:
## Motivation

Enable React Optimizer (React Compiler) for the published runtime and default theme, building on the React Hooks fixes merged in #3673. Rspress users receive precompiled components without configuring the optimizer in their applications.

Rebased onto the latest `main`, with the already-merged documentation and Hooks fixes removed from this PR.

## Changes

- Enable `reactCompiler` in both React library builds in `packages/core/rstack.config.ts`, using the native Rsbuild/SWC integration.
- Target React 18 and add `react-compiler-runtime` as a direct dependency to support React 18 and 19, following the [React library compilation guidance](https://react.dev/reference/react-compiler/compiling-libraries).
- Retain the stable default `Content` fallback and remove the imported function from the `PackageManagerTabs` memo dependencies.

The changes preserve the original author's commits. Compiler integration uses the existing build toolchain; no Babel plugin is added.

## Validation

- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm build`
- `pnpm check` — lint, type checking, and formatting passed
- `pnpm test:unit` — 383 passed, 1 skipped
- `pnpm test:e2e` — all 226 tests passed across 72 files, including React 18/19, Markdown generation, tabs, navigation, and hydration-related fixtures
- Verified compiler-runtime imports in emitted runtime and theme code, including `Content`, `Badge`, and `Tabs`.


sanjaiyan-dev:
To provide additional context on why shipping React Compiler with Rspress is a safe and high-impact improvement, here is how other major production applications and libraries have benefited from it:

### 📊 Real-World Production Benchmarks

* **[Meta Quest Store](https://react.dev/blog/2025/10/07/react-compiler-1#react-compiler-at-meta)**:
  * **12% improvement** in initial load and cross-page navigation.
  * **Up to 2.5× faster** interaction speeds while keeping memory neutral.

* **[Wakelet (100% User Rollout)](https://github.com/reactwg/react-compiler/discussions/52)**:
  * **15% overall INP improvement** (275ms ➔ 240ms), with pure UI components seeing up to **~30% INP speedups**.
  * **10% LCP improvement** (2.6s ➔ 2.4s).
  * Described the rollout as a high-return, near *"free lunch"* improvement for core UX and Lighthouse metrics.

* **[Sanity Studio & Component Libraries](https://github.com/reactwg/react-compiler/discussions/33)**:
  * Successfully shipped compiled libraries targeting **React 18** using `react-compiler-runtime` ([`@sanity/ui`](https://github.com/sanity-io/ui/pull/1444), [`@portabletext/editor`](https://github.com/portabletext/editor/pull/319/)).
  * Measured a **20–30% increase in eFPS (frames per second)** during continuous typing and editing interactions simply by toggling the compiler on.

---

### 💡 Key Takeaways Beyond Performance

1. **Surfaces Hidden Bugs**: Both Wakelet and Sanity noted that the compiler’s strict analysis helps surface long-standing "cold case" bugs and data races that were previously masked by excessive component re-renders.
2. **Proven Library Distribution Model**: Sanity's rollout confirms that shipping compiled UI packages via `react-compiler-runtime` for React 18/19 compatibility works reliably in production npm ecosystems.

Integrating this into Rspress gives our documentation users these exact out-of-the-box rendering gains without requiring any manual setup on their end.

packages/core/src/runtime/Content.tsx
```diff
@@ -2,8 +2,14 @@ import { type ReactNode, Suspense, useMemo } from 'react';
 import { useLocation } from 'react-router-dom';
 import { pathnameToRouteService } from './route';
 
+const DefaultFallback = <></>;
+
 // TODO: fallback should be a loading spinner
-export const Content = ({ fallback = <></> }: { fallback?: ReactNode }) => {
+export const Content = ({
```
sanjaiyan-dev, line 8: This change was made to fix the following error from react-compiler.

> `(BuildHIR::node.lowerReorderableExpression)` Expression type `JSXFragment` cannot be safely reordered.

packages/core/src/theme/components/PackageManagerTabs/index.tsx
```diff
@@ -155,7 +155,7 @@ export function PackageManagerTabs({
 
   const Pre = useMemo(() => {
     return getCustomMDXComponent().pre;
-  }, [getCustomMDXComponent]);
```
sanjaiyan-dev, line 158: This change was made to fix the following error, which is a stable external function `getCustomMDXComponent`.

>Found 1 error:
> Error: Found extra memoization dependencies
> Extra dependencies can cause a value to update more often than it should, resulting in performance problems such as excessive renders or effects firing too often.

packages/core/src/theme/components/Tabs/index.tsx
```diff
@@ -116,6 +116,20 @@ export const Tabs = forwardRef(
       return getTabValuesFromChildren(children, values);
     }, [children, values]);
 
+    const defaultValueIndex =
+      defaultValue !== undefined
+        ? tabValues.findIndex(item => item.value === defaultValue)
```
sanjaiyan-dev, line 121: Move the calling of React hook top to resolve the React compiler error and to maintain React standards:

> Error: Hooks must always be called in a consistent order, and may not be called conditionally. See the Rules of Hooks (https://react.dev/warnings/invalid-hook-call-warning)

SoonIter:
Thanks for taking the time to contribute this to Rspress. However, benchmarks from other products do not demonstrate that this change provides a meaningful benefit to Rspress. This PR adds a runtime dependency, changes the build pipeline, and requires source refactors, so I think it needs Rspress-specific evidence to justify the trade-off.

Could you provide before-and-after measurements on representative Rspress sites and interactions—for example, render counts, INP or frame time, bundle size, and build time—and identify the concrete bottlenecks this PR improves? Without that data, I don't think we have enough evidence to merge this change.

SoonIter:
Bundle size comparison

https://github.com/web-infra-dev/rspress/actions/runs/31038294665?pr=3586#summary-92540812580

image

These bundle size changes are within an acceptable range.

SoonIter:
😄 This part of the bundle size seems to be a constant, causing similar newly initialized projects to be about 10% larger.

image

sanjaiyan-dev:
Hi @SoonIter,

Extremely sorry for the delay! Here is the benchmark comparison and analysis I gathered from the React DevTools Profiler with screenshot proof.

---

### 1. `PackageManagerTabs ` Component (Tab Switching & Re-renders) 

* **Without Memoization (Screenshot 2):**
  * **Commit Render Duration:** Consistently sitting at **`~13.6ms`** across repeated state updates (commits at 1.3s, 2.0s, 2.7s, 3.4s, 4.3s).
  * **Flamegraph Behavior:** Every tab switch triggers unnecessary cascading re-renders down through sub-items (`Bun`, `Yarn`, `BlogBanner`, `Anonymous` wrappers, etc.).

* **With `React.memo` (Screenshot 1 - marked with `Memo ✨`):**
  * **Commit Render Duration:** Drops to **`~1.8ms – 2.0ms`** (commits at 2.0s, 2.8s, 3.4s, 4.0s, 5.1s).
  * **Result:** **~86% reduction** in commit render time. Subtree re-rendering is skipped cleanly when switching active tabs.
  
  
Screenshot 2026-08-25 at 20 22 16
Screenshot 2026-08-25 at 20 26 27


---

### 2. `Overview` Page (Search / Filter Interaction)

* **Without Memoization (Screenshot 3):**
  * **Commit Render Duration:** Takes **`~12.1ms`** on every keystroke when typing into the filter box (`root`).
  * **Flamegraph Behavior:** Large waterfall where static UI parts re-render on every character typed (`LlmsCopyButton`, `H1`, `MyH1`, `Nav`, `SvgCopy`, and multiple SVG icons).

* **With `React.memo` (Screenshot 4 - marked with `Memo ✨`):**
  * **Commit Render Duration:** Down to **`~1.8ms – 2.0ms`** per filter update.
  * **Result:** **~85% speedup**. All unaffected header items, SVG icons, and copy buttons are bypassed during filtering, keeping keystroke input smooth with zero frame lag.
Generated Image August 25, 2026 - 9_03PM
Screenshot 2026-08-25 at 20 44 41

---

### Summary
* Both interactive paths (`PackageManagerTabs ` switching and `Overview` input filtering) went from **12ms–13.6ms down to ~1.8ms–2ms**.
* Avoids re-rendering static icons and non-dependent sibling trees.

Let me know what you think or if you'd like me to profile any other specific component before merging!

sanjaiyan-dev:
- Frame Budget: At 60fps (16.6ms budget) or 120fps (8.3ms), taking 13.6ms just for React render guarantees dropped frames during interactive operations.

- Filter input lag: In the Overview filter, this 12ms runs on every keystroke. Stacking keystrokes can easily block the main thread.

- Device scaling: On low-end hardware or 4x CPU throttle, that 13ms turns into 50ms+ (crossing the Long Task / INP threshold).

Dropping it to ~1.8ms leaves plenty of room on the main thread for layout and paint.

sanjaiyan-dev:
Added the profiler benchmark above (_~85%_ drop in render time on search/tab updates).

Alternatively, if we want to keep the code cleaner, we can opt into React Compiler with "use memo" annotation mode for these specific components. Happy to adjust either way :)

PR #3631, opened 2026-08-27
fix(theme): replace `border-bottom` link underlines with `text-decoration`

SoonIter:
## Summary

- replace `border-bottom` link underlines with `text-decoration` in document content, callouts, and prompt custom content
- apply a consistent `4px` underline offset across all three contexts
- preserve the existing visibility behavior: document links are underlined on hover, while callout and prompt links remain underlined by default
- ensure links containing inline code render the underline with the text instead of having it obscured by the code background

Validation:

- Prettier check for the three changed SCSS files
- `@rspress/core` build and Publint
- `git diff --check`

## Related Issue

N/A

## Screenshots

The visual change is limited to underline painting:

| Before | After |
| --- | --- |
| `border-bottom` can be covered by an inline code background | `text-decoration` follows the linked text with a consistent `4px` offset |

## Checklist

### DIFF
--- a/packages/core/src/theme/components/Callout/index.scss
+++ b/packages/core/src/theme/components/Callout/index.scss
@@ -111,7 +111,7 @@
     }
 
     a {
-      border-bottom: 1px solid currentColor;
+      text-decoration: underline;
     }
   }
 }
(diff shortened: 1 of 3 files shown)


SoonIter:
@codex review

Timeless0911 (approved):

PR #3638, opened 2026-08-30
fix(search): strip directive markers from search index content

buyfakett:
## Summary

This PR fixes local search snippets that exposed Markdown-only syntax such as container directive markers (`:::`) and GitHub alert markers (`[!TIP]`).

The search index now strips those marker lines while keeping the visible body text searchable. It also handles custom heading IDs and indented directive blocks parsed as searchable code content, while preserving real inline `:::` strings inside code.

### Before

image

### After

image

## Checklist

### DIFF
--- a/packages/core/src/node/route/extractPageData.test.ts
+++ b/packages/core/src/node/route/extractPageData.test.ts
@@ -597,6 +597,125 @@ describe('getPageIndexInfoByRoute', async () => {
     );
   });
 
+  it('should strip container directive markers from search content', async () => {
+    const pageIndexInfo = await getPageIndexInfoByRoute(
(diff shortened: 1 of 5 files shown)


buyfakett:
done

SoonIter (approved):
LGTM, thanks 

Rspress currently uses a dedicated pipeline, implemented in `extractPageData`, to strip MDX syntax and extract plain text.

MDX -> Text

However, this pipeline doesn't fully support all syntax extensions of MDX/Markdown. 

To address search indexing issues more systematically, we should improve this approach so we don't have to add regex-based workarounds for each unsupported syntax extension.

related to https://github.com/web-infra-dev/rspress/pull/3543

Issue #1725, opened 2024-12-31
[Enhancement]:Make sure that the search index is loaded successfully


Timeless0911:
### What problem does this feature solve?

![Image](https://github.com/user-attachments/assets/74c6f782-5394-4e02-8e36-b4d460921a4a)

### What does the proposed API look like?

- Search after json loading
- Give loading prompt

Timeless0911:
https://github.com/web-infra-dev/rspress/pull/1811

chenjiahan:
I think https://github.com/web-infra-dev/rspress/pull/1811 can not solve this problem completely, imagine the network is really bad...

PR #3678, opened 2026-09-13
fix(theme/Search): await FlexSearch addAsync before init resolves

### LINKED ISSUE #3658: [Bug]: LocalProvider.init() does not await FlexSearch addAsync, so the first search query can report "No matching results"
### Version

`@rspress/core` 2.0.21 (also present on `main`: `packages/core/src/theme/components/Search/logic/providers/LocalProvider.ts`).

### Details

`LocalProvider.init()` adds every page to the three FlexSearch `Document` indexes with `addAsync(...)` but never awaits the returned promises:

```ts
for (const item of pagesForSearch) {
  const flexSearchItem = item as unknown as FlexSearchCompatibleData;
  this.#index!.addAsync(item.routePath, flexSearchItem);
  this.#cjkIndex!.addAsync(item.routePath, flexSearchItem);
  this.#cyrillicIndex!.addAsync(item.routePath, flexSearchItem);
}
```

`SearchPanel.initSearch()` does `await searcher.init(); setInitStatus('inited')`, so `initStatus` becomes `'inited'` while FlexSearch is still indexing. With `tokenize: 'full'` on a medium index (our site: 952 documents per locale, 1.7 MB raw) indexing takes 1.7–2.3 s after the fetch, and `SearchPanel` renders `NoSearchResult` as soon as the suggestion list is empty and `initStatus === 'inited'` — no spinner. A query typed shortly after opening the panel therefore shows "No matching results" (7 of 8 trials in our measurement) until the user retypes.

FlexSearch 0.8's `addAsync` returns a promise that resolves once the add has been applied, so the fix is to await them before returning from `init()`:

```ts
const pending: Promise<unknown>[] = [];
for (const item of pagesForSearch) {
  const flexSearchItem = item as unknown as FlexSearchCompatibleData;
  pending.push(
    this.#index!.addAsync(item.routePath, flexSearchItem),
    this.#cjkIndex!.addAsync(item.routePath, flexSearchItem),
    this.#cyrillicIndex!.addAsync(item.routePath, flexSearchItem),
  );
}
await Promise.all(pending);
```

With that change the panel keeps its loading state until the index is ready and then runs the pending query (`initSearch` already re-runs `searcher.match(query)` after `init()` resolves). We are carrying this as a `pnpm patch` in https://github.com/ScriptedAlchemy/agent-bundle (https://github.com/ScriptedAlchemy/agent-bundle/pull/599, `patches/@rspress__core@2.0.21.patch`) and would be glad to drop it once upstream awaits the adds.

### Reproduce

1. Build a site with a few hundred pages (or throttle the CPU), `rspress preview`.
2. Open search (⌘K) and type a term that exists in a page body within ~1 s of opening.
3. Observe "No matching results"; delete and retype the last character after ~2 s → results appear.

devtechedge:
## Motivation

`LocalProvider.init()` issues `addAsync()` on the three FlexSearch indexes for every page but never awaits the returned promises. `SearchPanel.initSearch()` awaits `init()` and sets `initStatus` to `'inited'` while indexing is still running, so on a medium site a query typed shortly after opening search can render "No matching results" until the user retypes. This reproduces with 952 documents per locale, as described in #3658.

## Changes

Collect the `addAsync()` promises for each page's three indexes into a pending array and `await Promise.all(pending)` before `init()` resolves. The panel therefore keeps its loading state until the index is ready and then runs the pending query. No other behavior changes.

Related to #1725
Fixes #3658

## Verification

- `tsc --noEmit` on `packages/core` passes.
- Repo build and diff checks pass.

### DIFF
--- a/packages/core/src/theme/components/Search/logic/providers/LocalProvider.test.ts
+++ b/packages/core/src/theme/components/Search/logic/providers/LocalProvider.test.ts
@@ -1,0 +1,78 @@
+import type { PageIndexInfo } from '@rspress/core';
+import { describe, expect, rstest, test } from '@rstest/core';
+
+interface MockAddRecord {
+  id: string;
+  settled: boolean;
+}
+
+interface FlexSearchMockState {
+  pendingAdds: MockAddRecord[];
+}
+
+// FlexSearch resolves addAsync through an internal async queue, so searching
+// immediately after unawaited adds can still find documents. To make the
+// contract observable, replace Document with a mock whose addAsync settles in
+// a later macrotask and records whether init() waited for it.
+rstest.mock('flexsearch', () => {
+  const globalState = globalThis as unknown as {
+    __flexSearchMockState?: FlexSearchMockState;
+  };
+  globalState.__flexSearchMockState ??= { pendingAdds: [] };
+
+  class Document {
+    addAsync(id: string): Promise<void> {
+      const state = globalState.__flexSearchMockState!;
+      const record: MockAddRecord = { id, settled: false };
+      state.pendingAdds.push(record);
+      return new Promise(resolve => {
+        setTimeout(() => {
+          record.settled = true;
+          resolve();
+        }, 0);
+      });
+    }
+
+    searchAsync(): Promise<unknown[]> {
+      return Promise.resolve([]);
+    }
+  }
+
+  return { Document };
+});
+
+rstest.mock('virtual-page-data', () => ({ searchIndexHash: {} }));
+
+import { LocalProvider } from './LocalProvider';
+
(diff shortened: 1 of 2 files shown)


SoonIter (approved):
LGTM ♥️, thanks~

devtechedge:
Thanks for reviewing and merging **#3678**, and for maintaining **Rspress**.

This was an interesting race condition where `LocalProvider.init()` did not await FlexSearch `addAsync()` calls, allowing search initialization to finish before indexing was complete and causing intermittent empty results on larger sites.

I’m glad we could address it with a focused fix while preserving existing behavior.

Thanks again for the review and for the opportunity to contribute to Rspress!

PR #3681, opened 2026-09-15
fix(theme/Search): use native prefetch link tag for search index

SoonIter:
## Summary

- Add a prefetch link when `SearchPanel` renders so the current page's search index can start loading earlier.
- Reuse the search index URL resolution between prefetching and local search, limited to the current language/version index.
- Skip the link when search is disabled or no index exists for the current page.

closes #3680 

## Tests

- `pnpm exec rs check --type-check packages/core/src/theme/components/Search/SearchPanel.tsx packages/core/src/theme/components/Search/logic/providers/LocalProvider.ts`
- `pnpm exec rstest run --config rstest.e2e.config.mts e2e/fixtures/search/index.test.ts`


SoonIter:
“Lazy loading is for performance, while preloading is for user experience.”

In the search case, we hope to enhance the user experience through preloading not lazy loading

Timeless0911 (approved):

